PR #66, opened 2024-05-08
Embed test files

samarth-aws:
*Description of changes:*
Tests for the different analyses perform I/O to read from files which can cause errors and prevent the tests from running in parallel.

This PR avoids performing I/O in tests by embedding the test files into the test binaries via the `embed` package.

Notable changes:
- Move test files for each package to the package's own "testdata" directory, instead of everything residing in "./testdata/src/". This is necessary because of a limitation in the `embed` package.
- The backtrace tests that require files from the taint analysis are now in the `taint_test` package
- Refactor `analysistest.LoadTest` to use the `fs.FS` interface instead of performing file I/O directly
- Avoiding file I/O in tests allows them to be run in parallel
- Remove `go.mod` files from all the test directories because these files cannot be embedded via the `embed` package


victornicolet:
I see that it moves all testdata files into their respective packages, but then backtrace_test now needs to be in the taint package. That doesn't seem to be optimal to me?

samarth-aws:
> I see that it moves all testdata files into their respective packages, but then backtrace_test now needs to be in the taint package. That doesn't seem to be optimal to me?

It's not ideal but the `embed` package only supports embedding files in the same directory which is why I couldn't keep the existing `testdata/src/<package>` style. We could maybe have makefile rules (e.g. `make taint-test`) to run the tests in the `taint` package excluding the backtrace tests?

An alternative is to copy over the `taint/testdata` directory to the backtrace directory, but that also seems sub-optimal to me.

victornicolet:
@amzn-jasonrk do you have an opinion on this?
I was hoping the change to embedding would not change the testing files structures that is clearly separate from the source code (and makes working with complex tests in an IDE easier).

amzn-jasonrk:
Is this the only way to fix the flaky tests? It seems like there is something deeper that we should fix, especially if this is only happening on one machine. Have we confirmed the root cause analysis of the bug?

analysis/config/config.go
```diff
@@ -371,6 +367,32 @@ func Load(filename string) (*Config, error) {
 	return cfg, nil
 }
 
+// LoadEscape adds the escape configuration settings from escapeConfigBytes into c.
+func LoadEscape(c *Config, escapeConfigBytes []byte) error {
```
victornicolet, line 387: It's a little counterintuitive to have the user of the package manage the loading of different parts of the configuration. Does the user have to retrieve the escape config file info, load it, and then call this function which will do the check again before using the bytes? 
Can we provide a function that does all the loading?

analysis/taint/taint_utils_test.go
```diff
@@ -236,47 +239,174 @@ func expectTaintedCondInFuncs(funcNames ...string) func(error) bool {
...
+	escape2source := make(analysistest.TargetToSources)
+	type sourceInfo struct {
+		meta string
+		pos  token.Position
 	}
-	sink2source, _ := analysistest.GetExpectedTargetToSources(dir, ".")
-	for sink, sources := range sink2source {
-		for source := range sources {
-			fmt.Printf("Source %s -> sink %s\n", formatutil.SanitizeRepr(source), formatutil.SanitizeRepr(sink))
+	sourceIDToSource := map[string]sourceInfo{}
```
victornicolet, line 335: Was this code moved from analysistest to here because it's only used in taint tests now?

analysis/taint/testdata/basic/config.yaml
```diff
@@ -1,20 +1,20 @@
 taint-tracking-problems:
  - # The package regex matches all possible ways the package name might appear depending on how the program is loaded
   sources:
-    - package: "(main)|(command-line-arguments)"
+    - package: "(basic)|(main)|(command-line-arguments)"
```
victornicolet, line 4: Is that the name of the package that will be used when program is loaded in tests or just added for safety?

analysis/taint/testdata/builtins_121/config.yaml
```diff
@@ -2,13 +2,8 @@
 taint-tracking-problems:
   -
     sources:
-      - package: "(main)|(command-line-arguments)|(selects-test[[:graph:]]*)$"
+      - package: "(builtins_121)|(main)|(command-line-arguments)|(git.amazon.com[[:graph:]]*)$"
```
victornicolet, line 5: Looks like git is showing some wrong "file moves".

analysis/taint/testdata/implicit-flow/main.go
```diff
@@ -35,6 +35,7 @@ func example1() {
 		}
 	}
 
+	sink(data) // @Sink(ex1)
```
victornicolet, line 38: That's not because of a bug, just a new test correct?

samarth-aws:
> Is this the only way to fix the flaky tests? It seems like there is something deeper that we should fix, especially if this is only happening on one machine. Have we confirmed the root cause analysis of the bug?

I wouldn't say the tests are "flaky" since they fail consistently in Terminal.app but pass consistently in the GoLand terminal. I found the root cause of the issue: https://github.com/awslabs/ar-go-tools/issues/67

I still would like to merge this PR since doing file I/O in tests isn't great in general.

analysis/config/config.go
samarth-aws: `escapeConfigBytes` is the byte representation of the escape-config.json file. Used here: https://github.com/awslabs/ar-go-tools/pull/66/files#diff-8342ad4f71f577e5db7687522a945f15802f375a047ea3100c11d3ef2dfb442dR115

I can update the godoc to say this explicitly if that helps.

analysis/taint/taint_utils_test.go
samarth-aws: Yes, it's taint-analysis-specific and `analysistest` seems like a more general package.

analysis/taint/testdata/basic/config.yaml
samarth-aws: I think it's the name of the package when loaded via `packages.Load`. The tests fail without the config changes. Honestly I couldn't figure out where the name comes from or how to change it back to `main`, otherwise I would not have modified the config files.

analysis/taint/testdata/builtins_121/config.yaml
samarth-aws: That's weird, it's showing something similar for the selects-tests as well: https://github.com/awslabs/ar-go-tools/pull/66/files/15b61e6204d1b691aaf02d8c046d09f3ecf9c387#diff-629ad348bed03364019216de19f986a6be4f5dc2b67792e0278bc913fd863b21

analysis/taint/testdata/implicit-flow/main.go
samarth-aws: Yes it's a new test - I may have accidentally committed it from something else. I'm not sure how it got there but it can't hurt I guess.

analysis/config/config.go
samarth-aws: The idea is that `LoadX` functions shouldn't perform any I/O: the I/O should be the caller's responsibility. Maybe changing the type from `[]byte` to `io.Reader` could make this more obvious but I figured `[]byte` was the simpler option.

victornicolet: Maybe we can then add something in the return value of LoadConfig that tells the caller whether they need to do another LoadX? It should also not be the caller's responsibility to figure out the logic to load a complete config.

What I was proposing is another function that does I/O and is explicit about it. LoadConfigFromFile?

samarth-aws: True, maybe `config.Load` can take the config bytes and escape config bytes as arguments which would eliminate the need to have the `config.LoadEscape` function. Ideally the escape config would be part of the regular config file instead of a separate file.

`config.LoadGlobal` already performs file I/O. Granted, it is a special case but I'm not sure if we need a general-purpose `LoadConfigFromFile` function?

victornicolet: If we don't have a central place for the file I/O logic for loading a complete config, maintaining the cli which is harder to test will be difficult. I don't know if the cli was tested with the new functions (does the config reloading from disk work as expected?).

amzn-jasonrk (dismissed):
This is a big PR but it's mostly renaming and fixing up tests

victornicolet (dismissed):

amzn-jasonrk (dismissed):

victornicolet (dismissed):

amzn-jasonrk (approved):

victornicolet (approved):

PR #70, opened 2024-05-31
Fix non-determinism due to ignoring multiple target for bound labels

victornicolet:
*Description of changes:*
This fixes an issue identitified by @ArquintL where bound labels nodes with the same instruction but multiple targets (e.g. multiple free variable indices) were not properly tracked in the summary graph.
The added tests required fixing the backtrace analysis for a case where a closure binds a variable that is an argument of the enclosing function.

### DIFF
--- a/analysis/dataflow/function_summary_graph.go
+++ b/analysis/dataflow/function_summary_graph.go
@@ -74,7 +74,7 @@ type SummaryGraph struct {
 	SyntheticNodes map[ssa.Instruction]*SyntheticNode
 
 	// the label nodes of the function
-	BoundLabelNodes map[ssa.Instruction]*BoundLabelNode
+	BoundLabelNodes map[ssa.Instruction]map[BindingInfo]*BoundLabelNode
 
 	// the nodes accessing global information
 	AccessGlobalNodes map[ssa.Instruction]map[ssa.Value]*AccessGlobalNode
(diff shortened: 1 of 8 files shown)


analysis/dataflow/function_summary_graph.go
```diff
@@ -74,7 +74,7 @@ type SummaryGraph struct {
 	SyntheticNodes map[ssa.Instruction]*SyntheticNode
 
 	// the label nodes of the function
-	BoundLabelNodes map[ssa.Instruction]*BoundLabelNode
+	BoundLabelNodes map[ssa.Instruction]map[BindingInfo]*BoundLabelNode
```
samarth-aws, line 77: Since a `BoundLabelNode` already has a binding info field, would it be more efficient to have the type be `map[ssa.Instruction][]*BoundLabelNode` where the bound variable index determines the bound label node's position in the slice?

samarth-aws (approved):

ArquintL (approved):
LGTM

analysis/dataflow/function_summary_graph.go
victornicolet: The BoundLabelNode also has the closure information in it, I am not sure the stored value (at the instruction) is bound by only one closure here.

amzn-jasonrk (approved):

PR #71, opened 2024-06-05
Changing ExplicitFlowsOnly to FailOnImplicitFlow

victornicolet:
This changes the configuration option from ExplicitFlowsOnly to FailOnImplicitFlow which reflects better what the analysis does (fails on implicit flows when on) and by default does expliticit flows only, which is what taint analyzers usually do. 
Also makes previous applications using the .drawio config file format backwards compatible.

### DIFF
--- a/doc/01_taint.md
+++ b/doc/01_taint.md
@@ -122,12 +122,12 @@ The configuration contains some specific fields that allow users to tune how the
 
 #### Implicit flows
 Some flows can be implicit: for example, branching on tainted data. See the `implicit-flow` test for an example.
-By default, the taint analysis will produce an error if tainted data is branched on.
-If the tool should only track explicit taint flows, set the `explicit-flow-only` option to `true`:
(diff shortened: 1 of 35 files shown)


doc/01_taint.md
```diff
@@ -122,12 +122,12 @@ The configuration contains some specific fields that allow users to tune how the
 
 #### Implicit flows
 Some flows can be implicit: for example, branching on tainted data. See the `implicit-flow` test for an example.
-By default, the taint analysis will produce an error if tainted data is branched on.
-If the tool should only track explicit taint flows, set the `explicit-flow-only` option to `true`:
+By default, the taint analysis will not produce an error if tainted data is branched on.
+If the tool should fail when control flow depends on tainted data, set the `explicit-flow-only` option to `true`:
```
samarth-aws: The second clause of the sentence needs to be updated as well

samarth-aws (approved):

amzn-jasonrk (approved):

PR #74, opened 2024-06-21
Adding max-context option and calling max-depth unsafe.

victornicolet:
* Description of changes*
- Adds a new `max-entrypoint-context-size` option to limit the max context size at entrypoints.
- Change name of `max-depth` to `unsafe-max-depth` to make it explicit this option affects soundness.
- Set parsing of config file to fail on unknown fields to prevent misconfigurations.
- Cleanup config files in tests.


analysis/config/config.go
```diff
@@ -173,47 +165,41 @@ type StaticCommandsSpec struct {
 
 // Options holds the global options for analyses
 type Options struct {
```
victornicolet, line 167: Reorganized the fields in alphabetical order here because we have many.

analysis/config/testdata/full-config.yaml
```diff
@@ -37,15 +37,14 @@ options:
...
-  # short traces
-  max-depth: 42
-
   # this can be set to limit the number of alarms raised by an analysis if you only care about the truth value of
   # the output, or only want to analyze a few results
   max-alarms: 16
 
+  # this can be used to limit the size of the callstack when searching for contexts under which a function can be
+  # called.
+  max-context: 20
```
samarth-aws: Maybe naming it `max-calling-context-length` makes a bit more sense to distinguish it from other kinds of contexts?

analysis/config/config.go
```diff
@@ -222,26 +208,47 @@ type Options struct {
...
-	LogLevel int `xml:"log-level,attr" yaml:"log-level" json:"log-level"`
+	// SkipInterprocedural can be set to true to skip the interprocedural (inter-procedural analysis) step
+	SkipInterprocedural bool `xml:"skip-interprocedural,attr" yaml:"skip-interprocedural" json:"skip-interprocedural"`
 
 	// Suppress warnings
 	SilenceWarn bool `xml:"silence-warn,attr" json:"silence-warn" yaml:"silence-warn"`
+
+	// SourceTaintsArgs specifies whether calls to a source function also taints the argument. This is usually not
+	// the case, but might be useful for some users or for source functions that do not return anything.
+	SourceTaintsArgs bool `xml:"source-taints-args,attr" yaml:"source-taints-args" json:"source-taints-args"`
```
samarth-aws, line 232: We can leave this for a future PR but I think analysis-specific options should be part of that specific analysis' configuration, not the global options set

samarth-aws (dismissed):

analysis/config/testdata/full-config.yaml
victornicolet: I could call it `max-entrypoint-context-size` which specifies a little more where it applies? There's only one type of context so far, and for the full description you'd need to read the doc anyway (unless we give it a long name like `max-calling-context-length-for-entrypoint-disambiguation`)

analysis/config/config.go
victornicolet: Yes, I think once the options settle we should do some reorganization.

samarth-aws (approved):

amzn-jasonrk (approved):

PR #76, opened 2024-06-27
Adding pointer analysis configuration: no-effect functions and reflection

victornicolet:
Adds a new configuration option category, `pointer-config` and two options in the `pointer-config`:
- `reflection` sets using reflection to true in the pointer analysis when the analysis default state is intialized,
- `unsafe-no-effect-functions` allows the used to specify a set of functions for which no constraint is generated during the pointer analysis. This can dramatically help reducing false positive in some cases (see the `fmt.Errorf` example) but the user has to think about the soundness of the analysis when using the option.


analysis/dataflow/pointer.go
```diff
@@ -37,14 +38,20 @@ import (
 //
 // If error != nil, the *pointer.Result is such that every Value in the functions f such that functionFilter(f) is true
 // will be in the Queries or IndirectQueries of the pointer.Result
-func DoPointerAnalysis(p *ssa.Program, functionFilter func(*ssa.Function) bool, functionSet map[*ssa.Function]bool) (*pointer.Result,
-	error) {
+func DoPointerAnalysis(c *config.Config, p *ssa.Program,
```
samarth-aws, line 41: nit: `c` could have type `*config.PointerConfig` since it doesn't use all the fields of `*config.Config`

doc/01_taint.md
```diff
@@ -538,3 +537,18 @@ Note that the functions are identified by their fully qualified names. The full
...
+
+To solve that, the user can specify that the `fmt.Errorf` function should not generate any constraints during the pointer analysis:
+```yaml
+pointer-config:
+  unsafe-no-effect-functions:
+    - "fmt.Errorf"
+```
+All the functions whose full name (the name returned by `(*ssa.Function).String()`) matches one of the names provided by the user in the list of `unsafe-no-effect-functions` will be skipped during the pointer analysis (generating no constraints). If the user can prove that this is the case (i.e. no aliases are created by this function), then this is sound.
+
+>⚠️ Soundness of the analysis when `unsafe-no-effect-functions` are specified is the user's responsibility. One many also want to trade soundness for precision and specify effect-free functions for testing code properties.
```
samarth-aws: typo: "many" -> "may". I'm not sure what the second sentence is trying to say?

internal/pointer/summaries.go
```diff
@@ -0,0 +1,32 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
samarth-aws, line 1: I'm not sure how different licenses work within the same codebase so we may need to remove our copyright header here

analysis/dataflow/pointer.go
victornicolet: It could but I think having only one type of config throughout just makes things a little easier to read and refactor (if in the future we want to share more config options between our config and the pointer's config, e.g. log-level).
(Which also reminds me we were talking about using slog at some point)

doc/01_taint.md
victornicolet: Yes I need to rephrase that 😄  I want to say that developers may sometimes want to use our tool even when it's unsound. When it's unsound, it still useful to "test" some code properties (as opposed to prove).

internal/pointer/summaries.go
victornicolet: I think since this a a new file this is fine. We should include the BSD license, maybe in the pointer folder though (for the Go authors) and include "modified by ..." when we modify the pointer files. I'm not sure how to do the latter yet.

doc/01_taint.md
samarth-aws: True, that's a good point to mention

samarth-aws (approved):

amzn-jasonrk (approved):

PR #78, opened 2024-07-03
Fixes in tuple handling + output format improvements

victornicolet:
This PR improves tuple handling in the dataflow analysis:
- encoding of "non-tuples" explicit instead of through negative integers.
- tuple-manipulating instructions that are not at the boundary of function calls (returns, calls) now ignore tuple indices.
- analysis panics explicitly if indices don't match.
- some changes in the output format of the taint analysis to display type of nodes and make position more visually obvious.

### DIFF
--- a/analysis/dataflow/function_summary_graph.go
+++ b/analysis/dataflow/function_summary_graph.go
@@ -633,7 +633,6 @@ func (g *SummaryGraph) addEdge(source MarkWithAccessPath, dest GraphNode, cond *
 					updateEdgeInfo(source, dest, cond, sourceNode)
 				}
 			}
-
 		}
 	}
 
@@ -698,7 +697,6 @@ func isDiffNode(mark MarkWithAccessPath, source GraphNode, dest GraphNode) bool 
 
 func updateEdgeInfo(source MarkWithAccessPath, dest GraphNode, info *ConditionInfo, sourceNode GraphNode) {
 	outMap := sourceNode.Out()
-	index := source.Mark.Index
 	relPath := map[string]map[string]bool{source.Mark.Label: {source.AccessPath: true}}
 	edgeInfos, destPresent := outMap[dest]
 	if !destPresent {
@@ -707,7 +705,7 @@ func updateEdgeInfo(source MarkWithAccessPath, dest GraphNode, info *ConditionIn
 	edgeInfoFound := false
 	for _, edgeInfo := range edgeInfos {
 		// add the access path to each edge with matching index
-		if edgeInfo.Index == index {
+		if edgeInfo.Index == source.Mark.Index.Value {
 			edgeInfoFound = true
 			if _, ok := edgeInfo.RelPath[source.Mark.Label]; ok {
 				edgeInfo.RelPath[source.Mark.Label][source.AccessPath] = true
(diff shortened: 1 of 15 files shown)


analysis/dataflow/mark.go
```diff
@@ -49,6 +49,17 @@ const (
 	If
 )
 
+// IndexKind qualifies the kind of index of mark. It is either the index on a value returned by a function,
+// or not an actual index.
+type IndexKind int
+
+const (
+	// ReturnedTupleIndex is  index on a tuple returned by a functoin
```
samarth-aws: nit: typos

analysis/dataflow/function_summary_graph.go
```diff
@@ -1295,10 +1292,14 @@ func ppEdge(w io.Writer, n GraphNode, c EdgeInfo, arrow string) {
...
-	if len(c.RelPath) > 0 {
-		prefix += "@"
-		for inPath, outPaths := range c.RelPath {
-			for outPath := range outPaths {
+	prefixed := false
+	for inPath, outPaths := range c.RelPath {
+		for outPath := range outPaths {
+			if inPath != "" || outPath != "" {
+				if !prefixed {
+					prefix += "@"
```
samarth-aws, line 1300: Doesn't need to be addressed in this PR but I'm wondering if we should come up with a more efficient representation for paths? Strings are relatively expensive to update because they are immutable. Maybe it would be more efficient to use `[][]byte` or `[]string` instead?

analysis/dataflow/mark.go
```diff
@@ -49,6 +49,17 @@ const (
...
+	ReturnedTupleIndex IndexKind = iota
+	// NonIndex is a non-index
+	NonIndex
```
samarth-aws: Shouldn't `NonIndex` be the default value (0)?

samarth-aws (dismissed):

analysis/dataflow/mark.go
victornicolet: I don't think I use the default value, but that's a good point, it should be.

analysis/dataflow/function_summary_graph.go
victornicolet: We could, the limitation is that we need to match the pointer analysis' representation. Currently there is a 1-1 mapping between the two, they are both strings and have the same "syntax". Now that we have the pointer analysis internally, we can change that representation. We could also add some parsing step to translate between representations.

samarth-aws (approved):

analysis/dataflow/mark.go
```diff
@@ -76,6 +87,22 @@ func (m MarkType) String() string {
 	}
 }
 
+// MarkIndex wraps an index kind and the index value together
+type MarkIndex struct {
+	// Kind is the kind of index (currently, either a real index or not)
```
amzn-jasonrk, line 92: Do we plan to have other kinds of indices?

amzn-jasonrk (approved):

analysis/dataflow/mark.go
victornicolet: My thinking was that some point we might want to track indices resulting from `next` and `select` instructions as a special case.

PR #81, opened 2024-07-10
Adding annotation framework

victornicolet:
This PR adds an annotations package to manage code annotations.
Annotations are of the form:
```
//argot:<target specifier> <target specified args> <target annotations>
```
Target specifiers added are:
- `param` annotates a parameter,
- `function` annotates a function,
- `ignore` annotates a warning to ignore,
- `config` annotates a config

This allows to specify a taint analysis problem in the source code as opposed to the config file.


analysis/annotations/annotations.go
```diff
@@ -0,0 +1,378 @@
...
+const annotationPrefix = "//argot:"
+
+// AnnotationKind characterizes the kind of annotations that can be used in the program.
+type AnnotationKind = int
+
+const (
+	// Sink is the kind of Sink(...) annotations.
+	Sink AnnotationKind = iota
+	// Source is the kind of Source(...) annotations
+	Source AnnotationKind = iota
```
samarth-aws: Why all the `iota`s? Don't you just need one?

analysis/annotations/annotations.go
```diff
@@ -0,0 +1,378 @@
...
+	// SetOptions is the kind of SetOptions(...) annotations
+	SetOptions AnnotationKind = iota
+)
+
+// AnyTag is the special tag used to match any other tag
+const AnyTag = "_"
+
+// TargetSpecifier is the type of target specifiers in the annotations: an annotation is of the shape
+// "//argot:<target specifier> <target specifier arguments>* <annotations>
+type TargetSpecifier = string
```
samarth-aws, line 51: Probably don't need a type alias here since `TargetSpecifier` should be a distinct type from `string`

analysis/annotations/annotations.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+}
+
+// loadConfigTargetAnnotation loads a config annotation. Config annotations look like
+// "//argot:config tag SetOptions(option-name-1=value1,option-name-2=value2)" and are always linked to a specific problem
+// tag.
+func (pa ProgramAnnotations) loadConfigTargetAnnotation(logger *config.LogGroup, annotationContents []string, position string) {
+	if len(annotationContents) <= 2 {
+		logger.Warnf("argot:config expects a target tag and one or more SetOptions")
+		logger.Warnf("a comment is likely missing something at %s", position)
```
samarth-aws, line 307: missing return?

analysis/annotations/annotations.go
```diff
@@ -0,0 +1,378 @@
...
+		if param.Name() == paramName {
+			foundParam = true
+			if _, ok := annotations.params[param]; !ok {
+				annotations.params[param] = []Annotation{}
+			}
+			paramAnnotations, err := parseAnnotationContent(annotationContent)
+			if err != nil {
+				return err
+			}
+			annotations.params[param] = append(annotations.params[param], paramAnnotations...)
```
samarth-aws, line 352: nit: could `break` out of the loop here since the parameter has been found

analysis/annotations/annotations.go
```diff
@@ -0,0 +1,378 @@
...
+		idents := kindRegexp.FindStringSubmatch(contents)
+		if len(idents) > 1 {
+			parsedAnnotations = append(parsedAnnotations, Annotation{Kind: kind, Tags: parseAnnotationArgs(idents)})
+		}
+	}
+	return parsedAnnotations, nil
+}
+
+func parseAnnotationArgs(a []string) []string {
+	return funcutil.Map(strings.Split(a[1], ","), func(s string) string { return strings.TrimSpace(s) })
```
samarth-aws, line 377: potential index out of bounds

analysis/load_program.go
```diff
@@ -42,7 +42,7 @@ const PkgLoadMode = packages.NeedName |
 func LoadProgram(config *packages.Config,
 	platform string,
 	buildmode ssa.BuilderMode,
-	args []string) (*ssa.Program, error) {
+	args []string) (*ssa.Program, []*packages.Package, error) {
```
samarth-aws, line 45: nit: could make this return a `LoadedProgram` struct similar to `LoadedTestProgram`

analysis/taint/testdata/annotations/main.go
```diff
@@ -0,0 +1,87 @@
...
+	"strconv"
+	"strings"
+)
+
+// Specific problem-level config options can be set in annotations
+//argot:config lim SetOptions(max-alarms=2,unsafe-max-depth=2)
+
+// bar
+// It is also a source for the problem hybridDef also defined in the config.json
+// The sink for that problem is fmt.Println
```
samarth-aws, line 29: I feel like it could get confusing if sources and sinks are defined in different places. Maybe it won't be an issue if we have descriptive error messages

analysis/annotations/annotations.go
victornicolet: Yes, that's a mistake

analysis/annotations/annotations.go
victornicolet: The goal here is just readablity/documentation (and refactoring ease later if we need to make this a more complex type).

analysis/annotations/annotations.go
victornicolet: Yes, actually should refactor this to just return

analysis/taint/testdata/annotations/main.go
victornicolet: It might be confusing but I think it's necessary because you can't annotate third-party code, but you might want to annotate yours. 
I was thinking that we could add more information about "why was this identified as a sink/source/sanitizer/etc.." when doing the analysis, e.g. "this was a sink because it matched rule with "method": ..., "package": ...," or "matched an annotation ...".

analysis/annotations/annotations.go
samarth-aws: Agreed but that's possible with the syntax `type TargetSpecifier string` which I think is what you meant? Anyways this is more of a nit and a type alias mostly does the same thing

analysis/taint/testdata/annotations/main.go
samarth-aws: Also it would be nice if we could inform the user why something didn't match. Sometimes I'll run the analysis and it doesn't report anything because it couldn't find the entrypoints due to a typo in the config. So having Argot display something like the n "closest" matches to a config code-identifier would probably help the user debug the issue.

samarth-aws (dismissed):

amzn-jasonrk (dismissed):

victornicolet:
Rebased the branch after the backtrace output PR was merged.

amzn-jasonrk (approved):

samarth-aws (approved):

PR #83, opened 2024-07-11
Support individual field-sensitive functions

victornicolet:
*Description of changes:*
This adds support for individual field-sensitive functions (cherry-picked changes from diodon-wip branch).

### DIFF
--- a/analysis/config/config.go
+++ b/analysis/config/config.go
@@ -217,6 +217,8 @@ type Options struct {
 	//
 	// Note that the configuration option name is "field-sensitive" because this is the name that will be more
 	// recognizable for users.
+	//
+	// TODO deprecate since this case is covered by `"field-sensitive-funcs": [".*"]`?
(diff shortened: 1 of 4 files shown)


samarth-aws (approved):

analysis/config/config.go
```diff
@@ -518,6 +543,20 @@ func (c Config) MatchCoverageFilter(filename string) bool {
 	}
 }
 
+// IsPathSensitiveFunc returns true if funcName matches any regex in c.Options.PathSensitiveFuncs.
+func (c Config) IsPathSensitiveFunc(funcName string) bool {
```
amzn-jasonrk, line 547: Given the new annotations support, it might makes sense to unify the config so you can specify function-level attributes (such as path sensitivity, allowlisting, escape analysis type, etc via either an source annotation or a config regex. This would support both code we control as well as "annotating" system code.

amzn-jasonrk (approved):

PR #85, opened 2024-08-08
Udpate dependencies, fix logging and context for code identifiers.

victornicolet:
+ Improve logging and messages.
+ Small fix to avoid panic when calling Type on ssa values.
+ Add intermediate trace nodes for more understandable traces.
+ Stop analysis when max alarms is reached; allows faster debugging of the analysis.


samarth-aws (approved):

analysis/lang/values.go
```diff
@@ -236,3 +236,20 @@ func TryTupleIndexType(v types.Type, i int) types.Type {
 	}
 	return tupleType.At(i).Type()
 }
+
+// CanType checks some properties to ensure calling the Type() method on the value won't cause a sefgfault.
```
amzn-jasonrk, line 240: Is this a segfault (sefgfault?) or "just" a panic? I remember it crashing but not which kind.

amzn-jasonrk (approved):

analysis/lang/values.go
victornicolet: The error is something of the kind
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation
```
I.e. pacnicking because of a sfegfault (i'll correct the dumb typo in a future fix :D)

amzn-jasonrk: Interesting, it is both segfaulting and panicking. I would have thought it would be one or the other but maybe Go is doing some weird runtime magic with handling the segfault signal.

PR #89, opened 2024-09-11
Min go version 1.21 -> 1.22 due to vulnerability report, support go 1.23

victornicolet:
+ Bump up the minimum go version. `govulncheck` reports a vulnerability on the standard library of go1.21 that we can't upgrade.
+ Support go1.23
- Remove deadcode analysis in makefile. Deadcode tool fails with internal segfault.


Makefile
```diff
@@ -13,7 +13,7 @@ all: lint maypanic statistics reachability dependencies static-commands render t
 install: taint_install backtrace_install cli_install
 
 lint: **/*.go
-	deadcode -test -filter ar-go-tools/analysis ./...
+	# deadcode -test -filter ar-go-tools/analysis ./...
```
samarth-aws, line 16: Does deadcode no longer work with the new Go version?

analysis/dataflow/callgraph_122_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build go1.22
+//go:build !go1.22
```
samarth-aws: Why is this build tag disabled? It might be better to delete the file if its contents are duplicated in callgraph_123_test.go

.github/workflows/bat.yml
```diff
@@ -17,7 +17,7 @@ jobs:
   build:
     strategy:
       matrix:
-        go-version: ['1.21', '1.22' ]
+        go-version: ['1.22.7', '1.23']
```
samarth-aws, line 20: Why specifically 1.22.7? Wouldn't it be better to have 1.22 so we track the latest version?

Makefile
samarth-aws: A comment explaining why this is disabled would be useful. Maybe we can enable it in the future if the tool gets fixed

.github/workflows/bat.yml
victornicolet: Right now the CI will use go1.22.6 if 1.22.7 is not specified (same thing happened with go1.21 at some point). We will update later

analysis/dataflow/callgraph_122_test.go
victornicolet: I meant not 1.23, will fix.

samarth-aws (approved):

amzn-jasonrk (approved):

PR #91, opened 2024-09-17
Minor fixes and improvements in dataflow analyses and cli

victornicolet:
*Description of changes:*
- Changes when synthetic marks are created in the analysis: created at initialization instead of when visiting the instructions during intra-procedural analysis. As a results, marks are not propagated on creation to aliases and referrers like they were before and the results matches the intuition better: the sources are the location where a field is accessed (instead of all the aliases of a field access, which was confusing). 
- Eliminate an unused piece of code in pointer analysis and start investigating a bug when a type `struct{reflect.Value}` is used. 
- Add an option to skip bound labels in the analysis: dataflow-problem (taint, backtrace) level`unsafe-skip-bound-labels` option causes the analysis to ignore inter-procedural edges that follow aliases of bound variables in closures. This is *unsound*, but eliminates a lot of false positives and is reasonable in many cases: pointer information on aliases is still used for intra-procedural analysis. 
- Change config loading behavior in the cli, which now assumes a default config (provided by `config.NewDefault` instead of an empty config.
- A few other minor fixes:
   + terminal formatting when printing traces,
   + formatting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ lint: **/*.go
 test: **/*.go
 	go clean -testcache
 	go test ./analysis/...
+	go test ./internal/...
 
 argot-build: go.mod cmd/argot/**/*.go
 	go build -o bin/argot ./cmd/argot/main.go
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,7 @@ lint: **/*.go
 
 test: **/*.go
 	go clean -testcache
-	go test ./analysis/...
-	go test ./internal/...
+	go test ./...
 
 argot-build: go.mod cmd/argot/**/*.go
 	go build -o bin/argot ./cmd/argot/main.go
--- a/analysis/dataflow/intra_procedural_instruction_ops.go
+++ b/analysis/dataflow/intra_procedural_instruction_ops.go
@@ -67,10 +67,7 @@ func (state *IntraAnalysisState) DoDebugRef(*ssa.DebugRef) {
 
 // DoUnOp analyzes unary operations and checks the operator to see whether is a load or a channel receive
 func (state *IntraAnalysisState) DoUnOp(x *ssa.UnOp) {
-	switch x.Op {
-	case token.ARROW:
-		state.optionalSyntheticNode(x, x, x)
-	case token.MUL:
+	if x.Op == token.MUL {
 		transferCopy(state, x, x.X, x)
 	}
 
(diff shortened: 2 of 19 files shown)


analysis/dataflow/intra_procedural_instruction_ops.go
```diff
@@ -67,10 +67,7 @@ func (state *IntraAnalysisState) DoDebugRef(*ssa.DebugRef) {
 
 // DoUnOp analyzes unary operations and checks the operator to see whether is a load or a channel receive
 func (state *IntraAnalysisState) DoUnOp(x *ssa.UnOp) {
-	switch x.Op {
-	case token.ARROW:
-		state.optionalSyntheticNode(x, x, x)
```
samarth-aws, line 72: Don't we still need the `token.ARROW` case to ensure data-flow that data-flow propagates to the l-value of a channel receive expression in cases where the channel is not a source?

cmd/argot/cli/cli.go
```diff
@@ -118,7 +118,7 @@ func Run(flags tools.CommonFlags) {
 
 func seekConfig(configPath string, args []string) (*config.Config, bool, error) {
 	var err error
-	pConfig := &config.Config{} // empty default config
+	pConfig := config.NewDefault()
```
samarth-aws, line 121: Honestly if there's no config file, I'd rather have it be an error. Is there any case where a user would want the default config?

victornicolet: I typically use it to quickly inspect aliasing in a function, or see what the dataflow summary is. Usually it doesn't require setting any non-default option. I think it makes sense for the cli but less for the other tools.

analysis/dataflow/intra_procedural_instruction_ops.go
victornicolet: The change shouldn't affect the behaviour outside marking the synthetic nodes. On line 74 in that function the `simpleTransfer(state, x, x.X, x)` call is moving the marks from the channel `x.X` to the l-value `x`.

samarth-aws (dismissed):

Makefile
```diff
@@ -22,6 +22,7 @@ lint: **/*.go
 test: **/*.go
 	go clean -testcache
 	go test ./analysis/...
+	go test ./internal/...
```
samarth-aws: Maybe we can simplify this to `go test ./...` so we don't miss anything

samarth-aws (approved):

Issue #31, opened 2023-09-26
Improving callgraph precision by inlining


victornicolet:
Some analyzers improve the callgraph precision by "inlining" some calls before running the analyses:
https://github.com/google/capslock/blob/main/analyzer/rewrite.go#L99
We could use a similar configurable approach.

PR #94, opened 2024-10-02
Add a rewriting step to inline function calls in some interface consuming functions

victornicolet:
This adds a step before building SSA to "inline" some function calls as proposed in #31.
Currently, the calls to the functions passed to `slices.Sort`, `slices.SortStable` and `(*sync.Do).Once` are inlined in the function calling. We will add other functions as necessary, and possibly other sound rewrite steps. 

We may surface this to the user later if some options are worth it.


analysis/escape/escape_test.go
```diff
@@ -97,7 +97,7 @@ func TestSimpleEscape(t *testing.T) {
 	t.Parallel()
 
 	dirName := filepath.Join("./testdata", "simple-escape")
-	lp, err := analysistest.LoadTest(testfsys, dirName, []string{})
+	lp, err := analysistest.LoadTest(testfsys, dirName, []string{}, true)
```
samarth-aws: nit: Can we make `LoadTest` take an options struct as well? It's difficult to know what `true` refers to when reading

analysis/load_program.go
```diff
@@ -39,27 +40,39 @@ const PkgLoadMode = packages.NeedName |
...
+// LoadProgramOptions combines all the options that are used when loading programs.
+type LoadProgramOptions struct {
+	// BuildMode is the mode used when creating the SSA from the packages.
+	BuildMode ssa.BuilderMode
+	// LoadTests is a flag indicating whether tests should be loaded with the program.
+	LoadTests bool
+	// ApplyRewrites is a flag indicating whether the standard source rewrites should be applied.
+	ApplyRewrites bool
+	// Platform indicates which platform the analysis is being performed on (sets GOOS in env).
+	Platform string
```
samarth-aws, line 52: Should this be an enum (I guess the Go equivalent is constants with a custom type) to restrict the set of platforms to the ones Go supports? Maybe there's already one defined in the standard library or x/tools?

analysis/refactor/rewrite/rewriters_test.go
```diff
@@ -0,0 +1,111 @@
...
+				t.Fatalf("Incorrect call to %s", x)
+			}
+		}
+		for _, y := range expected {
+			seen[y] = seen[y] || strings.Contains(d.String(), y)
+		}
+
+	}
+
+	for _, y := range expected {
```
samarth-aws, line 111: Should the test fail if `expected` is empty?

analysis/load_program.go
victornicolet: Yes I can look into that.

analysis/refactor/rewrite/rewriters_test.go
victornicolet: I think it's acceptable to only scan for `unexpected` things and not fail if `expected` is empty? Is that what you meant?

samarth-aws: My bad, since `unexpected` and `expected` are parameters, there's no need to check if they're empty. I was thinking of a case where the test would trivially pass even if there's a bug in the program.

So in this case, if `len(state.PointerAnalysis.CallGraph.Nodes[i].Out) == 0`, then all the values of `seen` will be false and the test will pass, even though this would be unexpected behavior

victornicolet: I'll add a check for length

analysis/load_program.go
victornicolet: I didn't find a enum or a custom type that does it. There's a build-generated file that contains a map of valid GOOSes but its a private var.

samarth-aws (approved):

analysis/refactor/rewrite/rewriters.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// ApplyRewrites applies a series of rewrite steps to the packages and all its dependencies.
+// The rewrites are meant to preserve soundness of the analysis, under the condition that the focus points of the
+// analysis are not the entrypoint or endpoints of analyses.
+// The rewrites are:
+//
+// All the rewrites from the base rewrites used in capslock package:
+// - calls to [sort.Slice] and [sort.SliceStable] are removed and replaced by calls to the functions used in sorting.
+// - calls to (sync.Once).Do are replaced by a call to the method in the sync.Once object.
```
amzn-jasonrk, line 29: We have to a bit careful here because this rewrite removes the writes performed by e.g. `Once.Do` to internal data. If we are relying on the pointer analysis to track all writes and there is a once.do in some protected space, this rewrite would miss it. In theory incorrect pointer analysis results can propagate arbitrarily far but I'm not sure I could construct an example of actual unsoundness from these functions, but I haven't considered it very carefully.

A more obviously sound version would be to clone the methods in question, but that would be significantly more complicated to implement.

amzn-jasonrk (approved):

analysis/refactor/rewrite/rewriters.go
victornicolet: I think in general that is true, but I don't think we can construct such an example with `sync.Once` for dataflow analysis, unless those objects are specifically the ones being tracked, or `Do` is marked as a sink/source. A `sync.Once` has some private fields which are not aliased in the implementation and the client can't alias them.
I will probably add more of those cases and also reason about them.

PR #97, opened 2024-10-04
Adding warning for usage of unsound features when a function is being…

### LINKED ISSUE #95: Warn users when functions using reflect are encountered
We should warn users when argot needs to summarize a function that does reflection.
Currently, warnings are produced for:
- unbounded defers,
- recovers,
- flowing into goroutines
But we should also make sure that `unsafe` and `reflect` usage triggers warning,although we should be parsimonious in triggering warnings.

victornicolet:
The tool will now warn when Go features that are a threat to soundness are use in a function being analyzed by the dataflow analyses (see #95 ):

Screenshot 2024-10-04 at 1 52 24 PM
This is in addition to the existing warnings about possibly unbound defer stacks and data flowing through Go calls.
This feature is meant to help users understand what may threaten soundness, but we won't give guarantees about "a warning will be produced iff the analyzer encounters a feature that is a threat to soundness".

### DIFF
--- a/analysis/dataflow/report.go
+++ b/analysis/dataflow/report.go
@@ -185,24 +185,32 @@ func FindUnsoundFeatures(f *ssa.Function) UnsoundFeaturesMap {
 				// Only warn for implementations.
 				return
 			}
-			typStr := callCommon.Value.Type().String()
-			if strings.Contains(typStr, "unsafe") {
+
+			var pkg string
+			if f, okF := callCommon.Value.(*ssa.Function); okF {
+				pkg = lang.PackageNameFromFunction(f)
+			}
+			if strings.HasPrefix(pkg, "unsafe") {
 				unsafeUsages[iPos] = fmt.Sprintf("Calling %s from unsafe package.", callCommon.Value.Name())
 				return
 			}
-			if strings.Contains(typStr, "reflect") {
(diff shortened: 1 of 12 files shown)


analysis/dataflow/report.go
```diff
@@ -106,3 +109,108 @@ func (s *AnalyzerState) ReportSummaryNotConstructed(callSite *CallNode) {
...
+				return
+			}
+
+		case *ssa.Alloc:
+			typ := instr.Type().Underlying()
+			if strings.HasPrefix(typ.String(), "unsafe") {
+				unsafeUsages[iPos] = fmt.Sprintf("Allocating object of type %s", typ.String())
+				return
+			}
+			if strings.HasPrefix(typ.String(), "reflect") {
```
samarth-aws: nit: This `strings.HasPrefix` check is different from the `strings.Contains` check.

I'm also wondering if there's a more standard way to get the name of a package? Maybe via `go/packages.Package.PkgPath`? https://pkg.go.dev/golang.org/x/tools/go/packages#Package

samarth-aws (dismissed):

analysis/dataflow/report.go
victornicolet: Good point, we actually already has the functionality in the `lang` package.

samarth-aws (approved):

PR #105, opened 2024-10-30
Problem spec options

victornicolet:
This PR:
+ moves the `unsafe-max-depth`, `max-entrypoint-context-size` and `max-alarms` options into a sub-option group `analysis-options` for options that apply individually to each analysis problem.
+ adds the possibility for taint and slicing problems to redefine their own `analysis-options`.
+ adds logging when option values are being overriden by annotations or problem-specific options.
+ adds some errors capturing when an old configuration is being used.
+ adds a check forcing tag to be unique for analysis problems when there are multiple
- deprecates the `unsafe-ignore-non-summarized` option which is useless when using on-demand summarization.

Since the previous configuration files are incompatible, this bumps the version to 0.3.0.

### DIFF
--- a/analysis/backtrace/backtrace.go
+++ b/analysis/backtrace/backtrace.go
@@ -139,6 +139,18 @@ func Analyze(logger *config.LogGroup, cfg *config.Config, prog *ssa.Program, pkg
 	var errs []error
 	resTraces := make(map[df.GraphNode][]Trace)
 	for _, ps := range cfg.SlicingProblems {
+		// Number of alarms is problem specific, not global
+		state.ResetAlarms()
+
+		state.Logger.Infof("Analyzing slicing problem %s", ps.Tag)
+		// Set problem-specific options
+		prevOptions := state.Config.AnalysisProblemOptions
+
+		// Overriding options with problem-specific config
+		if ps.AnalysisProblemOptions != nil {
+			config.OverrideWithAnalysisOptions(state.Logger, state.Config, ps.AnalysisProblemOptions)
+		}
+
 		visitor := &Visitor{
 			SlicingSpec: &ps,
 			Traces:      make(map[df.GraphNode][]Trace),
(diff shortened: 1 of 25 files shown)


analysis/backtrace/backtrace.go
```diff
@@ -139,6 +139,18 @@ func Analyze(logger *config.LogGroup, cfg *config.Config, prog *ssa.Program, pkg
...
+		// Number of alarms is problem specific, not global
+		state.ResetAlarms()
+
+		state.Logger.Infof("Analyzing slicing problem %s", ps.Tag)
+		// Set problem-specific options
+		prevOptions := state.Config.AnalysisProblemOptions
+
+		// Overriding options with problem-specific config
+		if ps.AnalysisProblemOptions != nil {
+			config.OverrideWithAnalysisOptions(state.Logger, state.Config, ps.AnalysisProblemOptions)
```
samarth-aws, line 151: We should log a message when overriding an option, preferably at info level

analysis/config/config.go
```diff
@@ -134,6 +134,32 @@ type Config struct {
...
+	// there is some potential alarm-causing result.
+	MaxAlarms int `xml:"max-alarms,attr" yaml:"max-alarms" json:"max-alarms"`
+
+	// UnsafeMaxDepth sets a limit for the number of function call depth explored during the analysis.
+	// The default is -1, and any value less than 0 is safe: the analysis will be sound and explore call depth
+	// without bounds.
+	//
+	// Setting UnsafeMaxDepth to a limit larger than 0 will yield unsound results, but can be useful to use the tool
+	// as a checking mechanism. Limiting the call depth will usually yield fewer false positives.
+	UnsafeMaxDepth int `xml:"unsafe-max-depth,attr" yaml:"unsafe-max-depth" json:"unsafe-max-depth"`
```
samarth-aws, line 152: I don't know if this makes sense as a global option since it only applies to the dataflow analyses

analysis/backtrace/backtrace.go
victornicolet: Yes, the function `OverrideWithAnalysisOptions` does log using the `state.Logger`.

analysis/config/config.go
victornicolet: But it can be convenient to define it for all the dataflow analyses once; it we define other type of problems and it doesn't have any effect it should be fine.

PR #106, opened 2024-10-31
Syntactic struct initialization analysis

samarth-aws:
*Issue #, if available:*

*Description of changes:*
Adds a new syntactic analysis called `struct-init` that identifies zero-allocations and invalid writes to a struct

### DIFF
--- a/analysis/dataflow/state.go
+++ b/analysis/dataflow/state.go
@@ -406,6 +406,10 @@ func (s *AnalyzerState) PopulateBoundingInformation(verbose bool) error {
 // To compute reachable functions without main or init, use the CallGraphReachable function with
 // the appropriate callgraph information.
 func (s *AnalyzerState) ReachableFunctions() map[*ssa.Function]bool {
+	if len(s.reachableFunctions) > 0 {
+		return s.reachableFunctions
+	}
+
 	// Create reachability information using best available callgraph
 	if s.reachableFunctions == nil {
 		s.reachableFunctions = make(map[*ssa.Function]bool)
--- a/analysis/dataflow/state.go
+++ b/analysis/dataflow/state.go
@@ -406,10 +406,6 @@ func (s *AnalyzerState) PopulateBoundingInformation(verbose bool) error {
 // To compute reachable functions without main or init, use the CallGraphReachable function with
 // the appropriate callgraph information.
 func (s *AnalyzerState) ReachableFunctions() map[*ssa.Function]bool {
-	if len(s.reachableFunctions) > 0 {
-		return s.reachableFunctions
-	}
-
 	// Create reachability information using best available callgraph
 	if s.reachableFunctions == nil {
 		s.reachableFunctions = make(map[*ssa.Function]bool)
--- a/analysis/syntactic/structinit/structinit.go
+++ b/analysis/syntactic/structinit/structinit.go
@@ -90,27 +90,30 @@ func Analyze(cfg *config.Config, prog *ssa.Program, pkgs []*packages.Package) (A
 	res := AnalysisResult{InitInfos: make(map[*types.Named]InitInfo)}
 	specs := structInitSpecs(cfg)
 
-	allocs := allStructAllocs(fns, specs)
+	allocs, structToNamed := allStructAllocs(fns, specs)
 	infos, err := initInfos(state, allocs, specs)
 	res.InitInfos = infos
 	if err != nil {
 		return res, err
 	}
-	logger.Debugf("initInfos: %+v\n", res.InitInfos)
-	for _, info := range res.InitInfos {
-		logger.Debugf("fieldToConst:\n")
-		for f, c := range info.fieldToConst {
-			logger.Debugf("\t%v -> %v\n", f, c)
-		}
-	}
+	debug(logger, res, structToNamed)
 
 	for _, alloc := range allocs {
 		if isConfiguredZeroAlloc(res, alloc) {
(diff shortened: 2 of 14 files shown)


cmd/argot/syntactic/syntactic.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	prog, pkgs, err := analysis.LoadProgram(opts, flags.FlagSet.Args())
+	if err != nil {
+		return fmt.Errorf("failed to load program: %v", err)
+	}
+
+	for _, syn := range cfg.SyntacticProblems {
+		if len(syn.StructInitProblems) > 0 {
+			logger.Printf("starting struct init analysis...\n")
+			res, err := structinit.Analyze(cfg, prog, pkgs)
```
victornicolet, line 72: I think it would be better if all the analyses share one state as opposed to construct a new state every time, even though in this case it's not an expensive operation. Some things that are built on-demand (like reachable functions) can be used by multiple analyses, yet created only when the first analysis that needs it creates it.

analysis/syntactic/structinit/structinit.go
```diff
@@ -0,0 +1,552 @@
...
+				}
+			}
+		}
+	}
+
+	return nil, false
+}
+
+// ReportResults writes res to a string and returns true if the analysis should fail.
+func ReportResults(res AnalysisResult) (string, bool) {
```
victornicolet, line 619: Will ReportResults have usage outside of logging results? i.e. is not using a logger intentional?

analysis/syntactic/structinit/structinit.go
```diff
@@ -0,0 +1,552 @@
...
+			if summaries.IsStdPackageName(lang.PackageNameFromFunction(instr.Parent())) {
+				return
+			}
+
+			switch instr := instr.(type) {
+			case *ssa.Store:
+				pos := program.Fset.Position(instr.Pos())
+				if write, ok := isInvalidWrite(res, instr, pos); ok {
+					namedType := write.structTypes.named
+					logger.Infof("found invalid write of value %v (wanted %v) to struct field %v.%v at %v\n", write.write.Got, write.write.Want, namedType, write.fieldType.Name(), pos)
```
victornicolet: nit: long line

analysis/syntactic/structinit/structinit.go
```diff
@@ -0,0 +1,552 @@
...
+// that was converted to an interface.
+type alloced struct {
+	val   ssa.Value       // val is the allocated value.
+	typs  structTypes     // typs are the types of val.
+	instr ssa.Instruction // instr is the allocation instruction.
+}
+
+// allStructAllocs returns all the instructions in fns that can allocate a value.
+// An allocation instruction is not always explicit:
+// MakeInterface instructions can also "allocate" a value.
```
victornicolet: Could changeType also do the same?

analysis/syntactic/structinit/structinit.go
```diff
@@ -0,0 +1,552 @@
...
+}
+
+// fieldAddrsOfAlloc returns all the instructions that address a field or
+// sub-field of the struct allocated in alloc.
+func fieldAddrsOfAlloc(alloc *ssa.Alloc) map[*ssa.FieldAddr]struct{} {
+	instrs := alloc.Block().Instrs
+	fieldAddrs := make(map[*ssa.FieldAddr]struct{})
+	vals := map[ssa.Value]struct{}{alloc: {}}
+	for _, instr := range instrs {
+		if addr, ok := instr.(*ssa.FieldAddr); ok {
```
victornicolet: nit: for readability, you could invert if statements to reduce nesting

analysis/syntactic/structinit/structinit.go
samarth-aws: I wanted the function to be logger-agnostic since it's called from the CLI but I can change it to use the standard library logger if that would be better

analysis/dataflow/state.go
```diff
@@ -406,6 +406,10 @@ func (s *AnalyzerState) PopulateBoundingInformation(verbose bool) error {
 // To compute reachable functions without main or init, use the CallGraphReachable function with
 // the appropriate callgraph information.
 func (s *AnalyzerState) ReachableFunctions() map[*ssa.Function]bool {
+	if len(s.reachableFunctions) > 0 {
```
victornicolet: I think this is the cause of the failing test. The state will recompute reachability based on the "best available callgraph" which will vary depending on whether pointer information has been computed. Checking whether this has already computed needs to be a little more involved (we can store an enum telling which callgraph computation has been used, and recompute if new information is available).

victornicolet: I'd remove this optimization in this PR and we can work on it again later.

analysis/syntactic/structinit/structinit.go
victornicolet: That's a good point

analysis/syntactic/structinit/structinit.go
samarth-aws: Good catch! I added some tests and found that the analysis is unsound so I need to fix it

cmd/argot/syntactic/syntactic.go
samarth-aws: Agreed, I can make that refactor once we implement the batch analysis feature since we'll have to make pretty much the same change for all the other analyses

victornicolet (approved):

PR #107, opened 2024-11-01
Config and new options

victornicolet:
This PR refactors the config and adds new options:
- dataflow problems are grouped under `dataflow-problems` with `taint-tracking` and `slicing` listing the respective dataflow problems. Dataflow specific options are specified under `dataflow-problems`, such as `summarize-on-demand`, `field-sensitive-funcs`, `field-sensitive` and `user-specs` (which were dataflow-specs). 
- a new option `project-root` specifies which directory all the the paths in the config are relative to.
- added `severity` and `description` fields in dataflow problems. 
- groups some analysis options that can be overriden per-problem (`max-alarms`) into a subset `analysis-options`.

### DIFF
--- a/analysis/config/config.go
+++ b/analysis/config/config.go
@@ -162,6 +162,32 @@ type AnalysisProblemOptions struct {
 	MaxEntrypointContextSize int `xml:"max-entrypoint-context-size,attr" yaml:"max-entrypoint-context-size" json:"max-entrypoint-context-size"`
 }
 
+// AnalysisProblemOptions are the options that are specific to an analysis problem.
+type AnalysisProblemOptions struct {
+	// MaxAlarms sets a limit for the number of alarms reported by an analysis.  If MaxAlarms > 0, then at most
+	// MaxAlarms will be reported. Otherwise, if MaxAlarms <= 0, it is ignored.
+	//
+	// This setting does not affect soundness, since event with max-alarms:1, at least one path will be reported if
+	// there is some potential alarm-causing result.
+	MaxAlarms int `xml:"max-alarms,attr" yaml:"max-alarms" json:"max-alarms"`
+
+	// UnsafeMaxDepth sets a limit for the number of function call depth explored during the analysis.
+	// The default is -1, and any value less than 0 is safe: the analysis will be sound and explore call depth
+	// without bounds.
+	//
+	// Setting UnsafeMaxDepth to a limit larger than 0 will yield unsound results, but can be useful to use the tool
+	// as a checking mechanism. Limiting the call depth will usually yield fewer false positives.
+	UnsafeMaxDepth int `xml:"unsafe-max-depth,attr" yaml:"unsafe-max-depth" json:"unsafe-max-depth"`
+
+	// MaxEntrypointContextSize sets the maximum context (call stack) size used when searching for entry points with context.
+	// This only impacts precision of the returned results.
+	//
+	// If MaxEntrypointContextSize is < 0, it is ignored.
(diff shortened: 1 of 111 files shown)


analysis/config/config.go
```diff
@@ -303,62 +256,46 @@ type Options struct {
 	// the case, but might be useful for some users or for source functions that do not return anything.
 	SourceTaintsArgs bool `xml:"source-taints-args,attr" yaml:"source-taints-args" json:"source-taints-args"`
```
samarth-aws: I think this should be a taint analysis-only option

analysis/config/config.go
```diff
@@ -286,12 +242,9 @@ type Options struct {
...
-	PathSensitiveFuncs []string `xml:"field-sensitive-funcs" yaml:"field-sensitive-funcs" json:"field-sensitive-funcs"`
-
-	// pathSensitiveFuncsRegexes is a list of compiled regexes corresponding to PathSensitiveFuncs
-	pathSensitiveFuncsRegexes []*regexp.Regexp
+	// ProjectRoot specifies the root directory of the project. All other file names specified in the config file are
+	// relative to the root. If not specified, the root is assumed to be the directory of the config file.
+	ProjectRoot string `xml:"project-root,attr" yaml:"project-root" json:"project-root"`
 
 	// SkipInterprocedural can be set to true to skip the interprocedural (inter-procedural analysis) step
 	SkipInterprocedural bool `xml:"skip-interprocedural,attr" yaml:"skip-interprocedural" json:"skip-interprocedural"`
```
samarth-aws: We should probably change this to `UnsafeSkipInterprocedural` because it can lead to unsound results

analysis/config/dataflow_config.go
```diff
@@ -0,0 +1,197 @@
...
+// DataflowProblems defines all the dataflow (taint, slicing) problems in a config file.
+type DataflowProblems struct {
+	// PathSensitive is a boolean indicating whether the analysis should be run with access path sensitivity on
+	// (will change to include more filtering in the future)
+	//
+	// Note that the configuration option name is "field-sensitive" because this is the name that will be more
+	// recognizable for users.
+	//
+	// TODO deprecate since this case is covered by `"field-sensitive-funcs": [".*"]`?
+	PathSensitive bool `xml:"field-sensitive" yaml:"field-sensitive" json:"field-sensitive"`
```
samarth-aws: Should we go ahead and remove this?

doc/01_taint.md
```diff
@@ -186,21 +197,38 @@ The `unsafe-max-depth` parameter controls how deep the dataflow paths can be. By
...
 ```yaml
 options:
+  analysis-options:
     unsafe-max-depth: 10
 ```
+You can also set it for a specific analysis problem:
+```yaml
+dataflow-problems:
+  taint-tracking:
+    - analysis-options:
```
samarth-aws: Should we name this `override-analysis-options` to make the behavior more clear

analysis/config/config.go
victornicolet: It really doesn't produce any results, it was used just as a sanity check that summary building doesn't fail. I wonder if we should just remove it.

analysis/config/dataflow_config.go
victornicolet: Good point 👍

analysis/config/config.go
victornicolet: Removed it.

samarth-aws (approved):

PR #110, opened 2024-11-04
Analysis targets

victornicolet:
This PR introduces analysis targets: the target programs for the `taint` and `backtrace` tools can be defined in the config file. 
See for example the config file in `payload/selfcheck/config.yaml`. Two targets `argot` and `racerg` are defined in the `targets:` top-level option, and then the `targets` option in each analysis subproblem specifies which program that analysis applies to.

Additional work needs to be done to improve the reporting when multiple targets are analyzed.

### DIFF
--- a/doc/00_intro.md
+++ b/doc/00_intro.md
@@ -35,7 +35,7 @@ The following tools are included in Argot:
 - the `render` tool can be used to render various representations of the code, such as its [Static Single Assignment](https://en.wikipedia.org/wiki/Static_single-assignment_form) (SSA) form or its callgraph (see [Render Tool](10_render.md#render-tool)).
 - the `racerg` tool, an experimental tool for data race detection (See [RacerG](11_racerg.md#racerg-sound-and-scalable-static-data-race-detector-for-go)).
 
-These tools can be used by developers to better understand their code, through the analysis of higher-level representations. For example, one can understand how the code is organized by looking at the callgraph, which abstract away the code and retains only the caller/callee information. In general, we do not state guarantees about the correctness of these tools, except for the dataflow analyses (see [taint analysis](01_taint.md#taint-analysis) and [backtrace](02_backtrace.md#backtrace-analysis))). However, we believe the representations obtained for each of these tools are useful enough to aid programmers.
+These tools can be used by developers to better understand their code, through the analysis of higher-level representations. For example, one can understand how the code is organized by looking at the callgraph, which abstract away the code and retains only the caller/callee information. In general, we do not state guarantees about the correctness of these tools, except for the dataflow analyses (see [taint analysis](01_taint.md#taint-analysis) and [backtrace](02_backtrace.md#backtrace-analysis)). However, we believe the representations obtained for each of these tools are useful enough to aid programmers.
 
 ### Configuration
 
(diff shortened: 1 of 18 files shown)


doc/00_intro.md
```diff
@@ -57,3 +57,28 @@ options:
 ```
 
 > 📝 The tool accepts five different settings for the logging level: 1 for error logging, 2 for warnings, 3 for info, 4 for debugging information and 5 for tracing. Tracing should not be used on large programs.
+
+
+### Targets
+
+The `taint` and `backtrace` tools of Argot support running against specific targets that can be defined in the config file (as opposed to passed in the command line).
+
+The targets are define in `targets`, and each target must have a `name` and a set of `files`. The file paths are taken relatively to the directory where the config file is by default, or relatively to the project root when the `option` `project-root` is defined.
```
samarth-aws: typo: define -> defined

doc/00_intro.md
```diff
@@ -57,3 +57,28 @@ options:
...
+    files:
+      - "cmd/target1/main.go"
+  - name: "target2"
+    files:
+      - "cmd/target2/main.go"
+
+```
+This assumes the config file is in a directory next to the `cmd` directory that contains all the executables of the project.
+Analysis problems that can use specific targets will have a `targets: ["target1", "target2"]` option for example. Then each tool can be run with a single argument, the config file.
+If some file paths are specified on the command line, the targets are ignored. The `"all"` target is a special target that means the problem applies to every target.
```
samarth-aws: What happens if the targets in the config differ from the ones passed in the command line?

payload/selfcheck/config.yaml
```diff
@@ -42,4 +52,14 @@ dataflow-problems:
         - type: "^int$" # not enough data in an int to do terminal injection
         - type: "^bool$" # same for bool
         - type: "^float" # same for floats
+  slicing:
+    # Check that the arguments of MustCompile are regexes defined entirely statically (i.e. all data flowing to the
```
samarth-aws: This isn't exactly true since the backtrace analysis won't report any errors if there are data flows to non-constants. We'd need to add extra functionality in the analysis to do this

internal/funcutil/collections.go
```diff
@@ -160,6 +160,16 @@ func Contains[T comparable](a []T, x T) bool {
 	return Exists(a, func(y T) bool { return x == y })
 }
 
+// Set takes an array of elements and returns a set of those elements, represented as a map from
+// elements to empty struct
+func Set[T comparable](a []T) map[T]any {
```
samarth-aws: Why is the resulting type `map[T]any` instead of `map[T]struct{}`?

doc/00_intro.md
victornicolet: If some files are passed on the command line, then the tools behave like before: all the analyses run on the target provided on the command line (the targets in the config file have no effect).

payload/selfcheck/config.yaml
victornicolet: True, it's more like we check the results of the analysis to make sure that it's the case.

internal/funcutil/collections.go
victornicolet: It should be `map[T]struct{}`

doc/00_intro.md
samarth-aws: Should we error or display a warning if that happens? It might confuse the user otherwise

victornicolet: I think we still have to figure out how to nicely handle reporting for the different targets; maybe we can figure this out then.

The goal here was to make configuration files portable while also allowing for the new "targets", so there are two modes:
- provide just a config file and no command line targets: everything is defined in the config file, this is useful when you have a fixed project and one config file for the project.
- provide a config file _and_ command line targets: if there are targets, they are ignored (the tool runs mostly like before). This means a config file is portable and can be used in different projects without having to change anything in it.

samarth-aws (approved):

PR #112, opened 2024-11-11
Json reports for taint and backtrace

victornicolet:
This PR adds reporting of traces in json format in the reporting directory. Users can quickly scan the `overall-report-*.json` file to see a map from problem tag to individual reports and a count of findings per severity level.

For example, run:
`argot backtrace -config payload/selfcheck/config.yaml`
or the same command with the `taint` tool.

### DIFF
--- a/analysis/config/report.go
+++ b/analysis/config/report.go
@@ -1,0 +1,154 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/awslabs/ar-go-tools/internal/formatutil"
+)
+
+// ReportInfo contains the information in the main report of argot
+type ReportInfo struct {
+	CountBySeverity map[Severity]int
+
+	// Reports groups entries by tag
+	Reports map[string]ReportGroup
+}
+
+// A ReportGroup lists the report contents in each details file
+type ReportGroup struct {
+	Tool     string
+	Severity Severity
+	Details  []string
+}
+
+// ReportEntry is one entry in the report, with high-level information about the severity and what tool
+// generated it. Contents of the report should be stored in a separate content file.
+type ReportEntry struct {
+	Tool        string
+	ContentFile string
+	Severity    Severity
+}
+
+// A ReportDesc gathers content and metadata about a specific report entry.
+type ReportDesc struct {
+	Tool     string
+	Tag      string
+	Severity Severity
+	Content  any
+}
+
+// NewReport returns a new ReportInfo with initialized maps
+func NewReport() ReportInfo {
+	return ReportInfo{
+		CountBySeverity: map[Severity]int{},
+		Reports:         map[string]ReportGroup{},
+	}
+}
+
+// Merge merges the report elements of other into the receiver
+func (r *ReportInfo) Merge(other *ReportInfo) {
+	for sev, count := range other.CountBySeverity {
+		r.IncrementSevCount(sev, count)
+	}
+	for tag, reportGroup := range other.Reports {
+		if thisReportGroup, ok := r.Reports[tag]; ok {
+			thisReportGroup.Details = append(thisReportGroup.Details, r.Reports[tag].Details...)
+		} else {
+			r.Reports[tag] = reportGroup
+		}
+	}
+}
+
+// IncrementSevCount increments the count of the provided severity label
+func (r *ReportInfo) IncrementSevCount(sev Severity, count int) {
+	if prev, ok := r.CountBySeverity[sev]; ok {
+		r.CountBySeverity[sev] = prev + count
+	} else {
+		r.CountBySeverity[sev] = count
+	}
+}
+
+func (r *ReportInfo) addEntry(tag string, entry ReportEntry) {
+	if _, tagPresent := r.Reports[tag]; !tagPresent {
+		r.Reports[tag] = ReportGroup{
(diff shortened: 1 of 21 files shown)


analysis/config/report.go
```diff
@@ -0,0 +1,154 @@
...
+// generated it. Contents of the report should be stored in a separate content file.
+type ReportEntry struct {
+	Tool        string
+	ContentFile string
+	Severity    Severity
+}
+
+// A ReportDesc gathers content and metadata about a specific report entry.
+type ReportDesc struct {
+	Tool     string
```
samarth-aws: nit: `Tool` should probably be the Go equivalent of an enum

analysis/config/report.go
```diff
@@ -0,0 +1,154 @@
...
+		Details:  newDetails,
+	}
+}
+
+// AddEntry write the information in the ReportDesc to a new file in the ReportsDir, and adds a new entry into the
+// receiver.
+func (r *ReportInfo) AddEntry(logger *LogGroup, c *Config, report ReportDesc) {
+	r.IncrementSevCount(report.Severity, 1)
+	logger.Debugf("Write report for tool=%s, tag=%s, severity=%s", report.Tool, report.Tag, report.Severity)
+	tmp, err := os.CreateTemp(c.ReportsDir, report.Tag+"-report-*.json")
```
samarth-aws, line 109: I'd rather have these methods write to an `io.Writer` instead of a temporary file so we could easily extend this to print reports in a json format to stdout in the future

analysis/config/report.go
```diff
@@ -0,0 +1,154 @@
...
+		ContentFile: tmp.Name(),
+		Severity:    report.Severity,
+	})
+}
+
+// Dump writes a new report file in the ReportsDir of the config file.
+func (r *ReportInfo) Dump(logger *LogGroup, c *Config) {
+	tmp, err := os.CreateTemp(c.ReportsDir, "overall-report-*.json")
+	if err != nil {
+		logger.Errorf("Could not write final report")
```
samarth-aws: nit: `err` is not reported

analysis/config/severity.go
```diff
@@ -0,0 +1,54 @@
...
+// FromString returns some severity when the string is a valid severity, otherwise none
+func FromString(s string) funcutil.Optional[Severity] {
+	if IsValidSeverityStr(s) {
+		return funcutil.Some(Severity(s))
+	}
+	return funcutil.None[Severity]()
+}
+
+// ToString converts the severity to a string type
+func (s Severity) ToString() string {
```
samarth-aws: nit: Go has a builtin interface for this that you can implement instead: `fmt.Stringer`

analysis/taint/dataflow_visitor.go
```diff
@@ -155,6 +155,15 @@ func (v *Visitor) Visit(s *df.AnalyzerState, source df.NodeWithTrace) {
...
 				if v.taints.addNewPathCandidate(NewFlowNode(v.currentSource), NewFlowNode(cur.NodeWithTrace)) {
 					reportTaintFlow(s, v.currentSource, cur)
+					s.Report.AddEntry(s.Logger, s.Config, config.ReportDesc{
+						Tool:     "taint",
+						Tag:      v.taintSpec.Tag,
+						Severity: v.taintSpec.Severity,
+						Content:  report(s, v.currentSource, cur, v.taintSpec),
+					})
+					if v.taintSpec.Severity == config.Critical {
+						panic(fmt.Sprintf("taint flows to critical location (problem tag %s)", v.taintSpec.Tag))
```
samarth-aws, line 165: We may have to experiment a bit with this to see what the best user experience is. What if there are two critical severity locations right next to each other and argot has been running for 30 minutes? Then the user will lose all that time and still be missing taint flows. Maybe it's best not to panic for now

analysis/taint/report.go
```diff
@@ -78,55 +78,83 @@ func reportCoverage(coverage map[string]bool, coverageWriter io.StringWriter) {
 // reportTaintFlow reports a taint flow by writing to a file if the configuration has the ReportPaths flag set,
 // and writing in the logger.
 // The function checks the annotations to suppress reports where there are //argot:ignore annotations.
-func reportTaintFlow(c *dataflow.AnalyzerState, source dataflow.NodeWithTrace, sink *dataflow.VisitorNode) {
-	c.Logger.Infof(" !!!! TAINT FLOW !!!!")
-	c.Logger.Infof(" 💀 Sink reached at %s\n", formatutil.Red(sink.Node.Position(c)))
-	c.Logger.Infof(" Add new path from %s to %s <== \n",
+func reportTaintFlow(s *dataflow.AnalyzerState, source dataflow.NodeWithTrace, sink *dataflow.VisitorNode) {
```
samarth-aws: Shouldn't this take a `FlowReport` instead?

analysis/taint/report.go
```diff
@@ -78,55 +78,83 @@ func reportCoverage(coverage map[string]bool, coverageWriter io.StringWriter) {
...
+// A FlowReport contains the information we serialize about a taint flow: a tag, the source and the sink, and
+// the trace from source to sink.
+type FlowReport struct {
+	Tag    string
+	Source dataflow.ReportNodeInfo
+	Sink   dataflow.ReportNodeInfo
+	Trace  []dataflow.ReportNodeInfo
+}
+
+// report generates a json report for a specific taint flow
```
samarth-aws: nit: maybe this should be `newFlowReport` instead since it doesn't have anything to do with json

analysis/taint/dataflow_visitor.go
victornicolet: We'll have to adjust, and the best way to handle this would probably to have another process listen to the results. We also don't want the user to wait 2 hours if a critical problem was encountered in the first minutes; that critical problem should be resolved before proceeding.

analysis/config/report.go
victornicolet: This is only the printing functionality, all the json content is in the `report.Content`. The role of this function is to manage where the content goes, in this case manage file creation and save file name for the other report that will also be saved to a file. 
The content could be printed to stdout separately somewhere, or we could choose to manage the Report differently and put everything in one file. This method has the advantage that you can listen to filesystem changes and scan reports as soon as they're out.

analysis/taint/report.go
victornicolet: I need to rename this to `logTaintFlow` and update the doc, it doesn't write to a file nor interact with the state's report. It only prints the results in the logger.

samarth-aws (approved):

PR #113, opened 2024-11-18
Using targets and tags in struct init analysis.

victornicolet:
Makes the `syntactic` tool use the targets like `taint` and `backtrace`.

### DIFF
--- a/analysis/config/config.go
+++ b/analysis/config/config.go
@@ -126,7 +126,7 @@ type Config struct {
 	// StaticCommandsProblems lists the static commands problems
 	StaticCommandsProblems []StaticCommandsSpec `yaml:"static-commands-problems" json:"static-commands-problems"`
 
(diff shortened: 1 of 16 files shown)


analysis/config/config.go
```diff
@@ -126,7 +126,7 @@ type Config struct {
 	// StaticCommandsProblems lists the static commands problems
 	StaticCommandsProblems []StaticCommandsSpec `yaml:"static-commands-problems" json:"static-commands-problems"`
 
-	SyntacticProblems []SyntacticSpecs `yaml:"syntactic-problems" json:"syntactic-problems"`
+	SyntacticProblems SyntacticSpecs `yaml:"syntactic-problems" json:"syntactic-problems"`
```
victornicolet, line 129: I changed this to be only a `SyntacticSpecs` since `SyntacticSpecs` already contains lists of specs.

samarth-aws (approved):

PR #115, opened 2024-11-18
Run with tags

victornicolet:
This PR makes some improvements in using tags to organize your analyses:
- in taint, backtrace and syntactic, specify `-tag my-tag` on the command line to run only the problems with tag `my-tag`.
- struct-init analysis in syntactic tool now respects `//argot:ignore tag` instruction for problems tagged with `tag`. Try `argot syntactic -config payload/selfcheck/config.yaml`

### DIFF
--- a/analysis/analyzers.go
+++ b/analysis/analyzers.go
@@ -48,7 +48,7 @@ type IntraAnalysisParams struct {
 // RunIntraProceduralPass runs an intra-procedural analysis pass of program prog in parallel using numRoutines, using the
 // analyzer state. The args specify the intraprocedural analysis parameters.
 // RunIntraProceduralPass updates the summaries stored in the state's FlowGraph
-func RunIntraProceduralPass(state *dataflow.AnalyzerState, numRoutines int, args IntraAnalysisParams) {
+func RunIntraProceduralPass(state *dataflow.AnalyzerState, numRoutines int, args IntraAnalysisParams) { //argot:ignore df-intra-uses
 	state.Logger.Infof("Starting intra-procedural analysis ...")
 	start := time.Now()
 
--- a/cmd/argot/tools/tools.go
+++ b/cmd/argot/tools/tools.go
@@ -31,6 +31,7 @@ type UnparsedCommonFlags struct {
 	ConfigPath *string
 	Verbose    *bool
 	WithTest   *bool
+	Tag        *string
 }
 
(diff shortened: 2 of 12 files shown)


cmd/argot/tools/tools.go
```diff
@@ -127,37 +132,31 @@ func LoadConfig(configPath string) (*config.Config, error) {
...
-			}
-		}
-		return structInitTargets
+		return targets(c.SyntacticProblems.StructInitProblems, allTargets, tag)
 	default:
 		return allTargets
 	}
 }
+
+func targets[T config.TaggedSpec](problems []T, allTargets map[string][]string, tag string) map[string][]string {
```
samarth-aws, line 152: nit: I don't think you need generics for this since `config.TaggedSpec` is an interface?

analysis/analyzers.go
```diff
@@ -48,7 +48,7 @@ type IntraAnalysisParams struct {
 // RunIntraProceduralPass runs an intra-procedural analysis pass of program prog in parallel using numRoutines, using the
 // analyzer state. The args specify the intraprocedural analysis parameters.
 // RunIntraProceduralPass updates the summaries stored in the state's FlowGraph
-func RunIntraProceduralPass(state *dataflow.AnalyzerState, numRoutines int, args IntraAnalysisParams) {
+func RunIntraProceduralPass(state *dataflow.AnalyzerState, numRoutines int, args IntraAnalysisParams) { //argot:ignore df-intra-uses
```
samarth-aws, line 51: Just a thought: I think it would be nice to specify exactly what kind of analysis finding we're ignoring with a given argot:ignore annotation. For this case, I'm thinking something like `//argot:ignore:zero-alloc df-intra-uses`, assuming this is a zero-allocation finding

samarth-aws (approved):

cmd/argot/tools/tools.go
victornicolet: It seemed the most concise way for *slices* of interfaces, otherwise you need to convert explicitly slices of implementations to slices of interfaces by creating a new slice of interfaces.

analysis/analyzers.go
victornicolet: Problem tags are unique (enforced when config is validated), so there can only be one analysis finding type for a given tag. I think users can put the related analysis in the tag name for readability of the annotation, but adding another argument to the annotation in this case wouldn't give you any more precision.

PR #116, opened 2024-11-20
Fix bugs 2

victornicolet:
This PR fixes bugs in the code:
- possible nil-panics identified with [Nilaway](https://github.com/uber-go/nilaway). Adds Nilaway as a check on the argot executable.
- management of reports directory and "project-root": now the project root is computed as an absolute path and all other paths are taken relatively to this.

### DIFF
--- a/analysis/config/config.go
+++ b/analysis/config/config.go
@@ -403,6 +403,12 @@ func LoadFromFiles(configFileName string) (*Config, error) {
 //
 //gocyclo:ignore
 func Load(filename string, configBytes []byte) (*Config, error) {
+	// Get absolute path to config. Other files in config will be relative to where th config is.
+	pathToConfig := filename
+	if !path.IsAbs(filename) {
+		wd, _ := os.Getwd()
+		pathToConfig = path.Join(wd, pathToConfig)
+	}
 	cfg := NewDefault()
 	unmarshallingError := unmarshalConfig(configBytes, cfg)
 	if unmarshallingError != nil {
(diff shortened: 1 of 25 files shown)


analysis/config/config.go
```diff
@@ -403,6 +403,12 @@ func LoadFromFiles(configFileName string) (*Config, error) {
 //
 //gocyclo:ignore
 func Load(filename string, configBytes []byte) (*Config, error) {
+	// Get absolute path to config. Other files in config will be relative to where the config is.
+	pathToConfig := filename
+	if !path.IsAbs(filename) {
+		wd, _ := os.Getwd()
```
samarth-aws: Should probably return the error

analysis/config/config_test.go
```diff
@@ -226,7 +227,8 @@ func TestLoadVersionBefore_v0_3_0_Errors(t *testing.T) {
 
 func TestLoadWithReports(t *testing.T) {
 	c := NewDefault()
-	c.ReportsDir = "example-report"
+	wd, _ := os.Getwd()
+	c.ReportsDir = path.Join(wd, "testdata/example-report")
```
samarth-aws, line 231: I think `filepath.Join` is preferred over `path.Join` for better OS compatibility: https://pkg.go.dev/path/filepath#Join

analysis/load_program.go
```diff
@@ -149,6 +149,13 @@ func LoadTarget(
 	logger *config.LogGroup,
 	cfg *config.Config,
 	loadTests bool) (*dataflow.AnalyzerState, error) {
+	if name != "" {
+		// If it's a named target, need to change to project root's directory to properly load the target
+		err := os.Chdir(cfg.Root())
+		if err != nil {
+			return nil, err
```
samarth-aws: nit: missing error context

samarth-aws (approved):

PR #117, opened 2024-11-21
Must be static option in backtrace

victornicolet:
This PR add the `must-be-static` option in the slicing specs of the backtrace tool. 
For example:
```
dataflow-problems:
  slicing:
    - must-be-static: true
      backtracepoints:
        - package: regexp
           method: MustCompile 
```
Reports an error only when non-static data reaches a call to `regexp.MustCompile`.

This PR also changes the behavior of backtrace in general to report an error when any trace is detected. `must-be-static` is a constraint on which traces are reported; only the non-static ones. 

Using the tool without the `must-be-static` means all traces are reported, and almost guarantees that an error will be shown. This should not be a problem since backtrace without the must-be-static option should only be used as an exploration tool right now.

### DIFF
--- a/analysis/backtrace/backtrace.go
+++ b/analysis/backtrace/backtrace.go
@@ -74,7 +74,7 @@ func Analyze(state *df.AnalyzerState) (AnalysisResult, error) {
 
 		state.Logger.Infof("Analyzing slicing problem %s", ps.Tag)
 		if ps.MustBeStatic {
-			state.Logger.Infof("Will check that data flowing to backtrace points is statis.\n")
+			state.Logger.Infof("Will check that data flowing to backtrace points is static.\n")
 		}
 		// Set problem-specific options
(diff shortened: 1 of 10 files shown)


analysis/backtrace/backtrace.go
```diff
@@ -73,6 +73,9 @@ func Analyze(state *df.AnalyzerState) (AnalysisResult, error) {
 		state.ResetAlarms()
 
 		state.Logger.Infof("Analyzing slicing problem %s", ps.Tag)
+		if ps.MustBeStatic {
+			state.Logger.Infof("Will check that data flowing to backtrace points is statis.\n")
```
samarth-aws: typo: statis -> static

samarth-aws (dismissed):

samarth-aws (approved):

PR #118, opened 2024-11-22
State refactor pt1

victornicolet:
This splits the `dataflow.AnalyzerState` into different `State` structs that embed each other:
- the `config.State` contains the config and logger.
- the `loadprogram.State` additionally contains the information about a built SSA program.
- the `ptr.State` additionally contains the information about the pointer analysis result.
- the `dataflow.State` additionally contains information about dataflows.
Other analyzers (taint, backtrace, dependencies, syntactic, ...) can build only the state they need, and the type tells what analysis results are available.


analysis/load_program_test.go
```diff
@@ -60,14 +63,14 @@ func programLoadTest(t *testing.T, files []string) {
...
 		LoadTests:     false,
 		ApplyRewrites: true,
 		Platform:      "",
 		PackageConfig: nil,
 	}
-	pkgs, _, err := LoadProgram(loadOptions, files)
+	pkgs, _, err := loadprogram.Do(files, loadOptions)
 	if err != nil {
 		t.Fatalf("error loading packages: %s", err)
 	}
```
samarth-aws, line 79: The test should make sure that `len(pkgs) > 0`

analysis/load_program_test.go
samarth-aws, line 1: nit: Should this belong in the loadprogram package?

analysis/loadprogram/state.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/awslabs/ar-go-tools/analysis/summaries"
+	"github.com/awslabs/ar-go-tools/internal/analysisutil"
+	"golang.org/x/tools/go/callgraph"
+	"golang.org/x/tools/go/callgraph/cha"
+	"golang.org/x/tools/go/packages"
+	"golang.org/x/tools/go/ssa"
+)
+
+// A State is the base state for the analyses in Argot. Analyses that do not require whole-program analysis
+// should be built with the go tools analysis framework.
```
samarth-aws: nit: a link to https://pkg.go.dev/golang.org/x/tools/go/analysis would be useful here

analysis/loadprogram/state.go
```diff
@@ -0,0 +1,216 @@
...
+type State struct {
+	*config.State
```
samarth-aws: Is it possible to mutate `config.State`? If not, maybe this shouldn't be a pointer

analysis/loadprogram/state.go
```diff
@@ -0,0 +1,216 @@
...
+	Target string
+
+	// a callgraph computed using the cha analysis. Useful to boostrap the reachable functions
+	chaCallgraph       *callgraph.Graph
+	reachableFunctions map[*ssa.Function]bool
+
+	// Stored errors
+	numAlarms  atomic.Int32
+	errors     map[string][]error
+	errorMutex sync.Mutex
```
samarth-aws: nit: this could be a single struct for clarity

analysis/analyzers.go
samarth-aws, line 1: I think this file should be in the dataflow package since it's only applicable to analyses that use a `dataflow.State`

analysis/backtrace/backtrace_taint_test.go
```diff
@@ -102,14 +100,13 @@ func runBacktraceTest(t *testing.T, test testDef, isOnDemand bool) {
...
 		t.Fatal("expect at least one taint tracking problem")
 	}
-	cfg.SlicingProblems = []config.SlicingSpec{{BacktracePoints: cfg.TaintTrackingProblems[0].Sinks}}
-	log := config.NewLogGroup(cfg)
-	state, err := dataflow.NewInitializedAnalyzerState(program, lp.Pkgs, log, cfg)
+	lp.Config.SlicingProblems = []config.SlicingSpec{{BacktracePoints: lp.Config.TaintTrackingProblems[0].Sinks}}
+	state, err := dataflow.NewDefault(lp.Config, lp.Prog, lp.Pkgs)
 	if err != nil {
-		t.Fatalf("failed to load state: %s", err)
+		t.Fatalf("failed to load datflow state: %s", err)
```
samarth-aws: nit: typo

analysis/backtrace/backtrace_taint_test.go
```diff
@@ -102,14 +100,13 @@ func runBacktraceTest(t *testing.T, test testDef, isOnDemand bool) {
...
 
-	if len(cfg.TaintTrackingProblems) < 1 {
+	if len(lp.Config.TaintTrackingProblems) < 1 {
 		t.Fatal("expect at least one taint tracking problem")
 	}
-	cfg.SlicingProblems = []config.SlicingSpec{{BacktracePoints: cfg.TaintTrackingProblems[0].Sinks}}
-	log := config.NewLogGroup(cfg)
-	state, err := dataflow.NewInitializedAnalyzerState(program, lp.Pkgs, log, cfg)
+	lp.Config.SlicingProblems = []config.SlicingSpec{{BacktracePoints: lp.Config.TaintTrackingProblems[0].Sinks}}
+	state, err := dataflow.NewDefault(lp.Config, lp.Prog, lp.Pkgs)
```
samarth-aws: I think `dataflow.NewDefaultState` would be a more specific name

analysis/config/state.go
```diff
@@ -0,0 +1,45 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+// ConfigLogger groups a config and a logger. All "state" structs should implement this.
+type ConfigLogger interface {
```
samarth-aws: I think `Configurer` would be a more fitting name in case we want to add a method in the future

analysis/builders.go
```diff
@@ -0,0 +1,85 @@
...
+// BuildWholeProgramTarget loads the target specified by the list of files provided. Return an analyzer state that has
+// been initialized with the program if successful. The Target of that state will be set to the provided name.
+func BuildWholeProgramTarget(
+	c *config.State,
+	name string,
+	patterns []string,
+	options loadprogram.Options) (*loadprogram.State, error) {
+	if name != "" {
+		// If it's a named target, need to change to project root's directory to properly load the target
+		err := os.Chdir(c.Config.Root())
```
samarth-aws: We should try to avoid these kinds of side-effects as much as possible. There's a way to load a program from a directory without cd-ing to it: https://github.com/awslabs/ar-go-tools/blob/mainline/internal/analysistest/analysistest.go#L59

analysis/builders.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/awslabs/ar-go-tools/analysis/config"
+	"github.com/awslabs/ar-go-tools/analysis/dataflow"
+	"github.com/awslabs/ar-go-tools/analysis/loadprogram"
+	"github.com/awslabs/ar-go-tools/analysis/ptr"
+	"github.com/awslabs/ar-go-tools/internal/formatutil"
+)
+
+// BuildWholeProgramTarget loads the target specified by the list of files provided. Return an analyzer state that has
+// been initialized with the program if successful. The Target of that state will be set to the provided name.
+func BuildWholeProgramTarget(
```
samarth-aws: It's confusing that there are two ways to construct a `loadprogram.State`: this function and `loadprogram.NewState`. I propose deleting all the `BuildXTarget` functions and making the `x.NewState` functions specify whatever dependencies they need as parameters.

analysis/backtrace/backtrace_taint_test.go
samarth-aws: Here's how I'd like this API to work using a dependency-injection style:
```
cfgState, err := config.NewState(...)
progState, err := analysistest.NewProgramState(cfgState, testfsys, ...)
ptrState, err := ptr.NewState(progState)
dfState, err := dataflow.NewState(ptrState)
res, err := backtrace.Analyze(dfState)
```

We may even be able to merge the implementation of `analysistest.LoadTest` into `loadprogram.NewState`, but that would mean that `loadprogram.NewState` would need a parameter that implements the filesystem interface which may be a bit much...

analysis/builders.go
victornicolet: Having tried to not have the `BuildXTarget` initially, I found myself just copy-pasting code everywhere that basically implements the `BuildXTarget`. 
The difference between the `NewState` functions for every state and the `BuildXTarget` is:
- with `BuildXTarget` you give files, load options and config and you get `X`. `BuildXTarget` does all the work.
- with `NewState` you have to build the program yourself, or the previous state yourself. `NewState` only builds the additional information on top of its arguments, so you have to chain `loadprogram.Do` and `NewState` for every state you want. This gives more control, but in most places (tests, executables) you just want `BuildXTarget`. `NewState` will be more useful when/if we don't specify a tool and `argot` just detects what analyses it needs to run and what states it needs to build.

analysis/builders.go
victornicolet: Good point.
Since this was a change from the previous PR and I'm not sure how well it will work with larger programs with vendored dependencies we can make that change in a future PR?

analysis/builders.go
samarth-aws: That's fair. How about renaming each `BuildXTarget` function to `<package>.NewDefaultState`?

analysis/backtrace/backtrace_taint_test.go
victornicolet: I think without a result monad the chaining of those operations where each returns an error becomes very verbose.
Also, one option would be to make the `config.State` contain all the arguments for the next `NewState`, i.e. loadprogram.NewState.

analysis/builders.go
victornicolet: There is still a difference that the `NewDefaultState` accepts an already built program as input (because it's really here only for tests) whereas the `BuildXTarget` takes in patterns as input, and then builds the program.

victornicolet: Ended up removing the `BuildXTarget` functions an instead introducing a result monad to chain the state building operations. Still have some cleanup to do, but this simplifies the interface.

analysis/dataflow/analyzers.go
```diff
@@ -12,8 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package analysis contains helper functions for running analysis passes.
-package analysis
+/*
```
samarth-aws, line 15: Redundant copyright header: I think `copyrights.sh` isn't smart enough to tell the difference between `//` and `/*` style comments

analysis/dataflow/state.go
```diff
@@ -82,90 +56,42 @@ type AnalyzerState struct {
...
 //   - linking aliases of bound variables to the closure that binds them
-func NewInitializedAnalyzerState(program *ssa.Program, pkgs []*packages.Package,
-	logger *config.LogGroup, config *config.Config) (*AnalyzerState, error) {
-	program.Build()
-	state, err := NewAnalyzerState(program, pkgs, logger, config, []func(*AnalyzerState){
-		func(s *AnalyzerState) { s.PopulateImplementations() },
-		func(s *AnalyzerState) { s.PopulatePointersVerbose(summaries.IsUserDefinedFunction) },
-		func(s *AnalyzerState) { s.PopulateGlobalsVerbose() },
+//
+// The State returned *does not* have dataflow information computed yet.
```
samarth-aws, line 68: Not necessary to implement now but it would be nice to have uninitialized state (maybe `dataflow.uninitializedState`) be a separate type from initialized state (`dataflow.State`), similar to the pattern I used for the CLI flags: https://github.com/awslabs/ar-go-tools/blob/mainline/cmd/argot/tools/tools.go#L29

Since we rarely compute all the dataflow information at once, it might get a little confusing but I still think it's worth having this distinction

analysis/dataflow/state.go
```diff
@@ -82,90 +56,42 @@ type AnalyzerState struct {
...
+func NewState(ps *ptr.State) result.Result[State] {
+	state, err := newFlowState(*ps, []func(*State){
```
samarth-aws: nit: Is there ever an instance where you don't populate the implementations, globals, and bounding information? I think this function can be simplified a lot by calling these functions explicitly instead of passing them in

internal/funcutil/result/result.go
```diff
@@ -0,0 +1,87 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package result
+
+import "fmt"
+
+// An Result holds a value or an error.
```
samarth-aws: nit: typo: "An" -> "A"

internal/funcutil/result/result.go
```diff
@@ -0,0 +1,87 @@
...
+	if v.IsOk() {
+		actual, _ := v.Value()
+		return f(actual)
+	}
+	_, actualErr := v.Value()
+	return err[S]{actualErr}
+
+}
+
+// Map is like hte map operation for the result monad
```
samarth-aws: nit: typo: "the"

internal/funcutil/result/result.go
```diff
@@ -0,0 +1,87 @@
...
+	if v.IsOk() {
+		actual, _ := v.Value()
+		return Ok[S](f(actual))
+	}
+	_, actualErr := v.Value()
+	return Err[S](actualErr)
+}
+
+// Do apply f to the content of the result if the result is ok
+func Do[T any](v Result[T], f func(*T)) {
```
samarth-aws, line 82: nit: to be more functional, should this return the result of applying `f` to `v` instead of mutating `v`?

internal/funcutil/result/result.go
```diff
@@ -0,0 +1,87 @@
...
+func Map[T any, S any](v Result[T], f func(*T) *S) Result[S] {
+	if v.IsOk() {
+		actual, _ := v.Value()
+		return Ok[S](f(actual))
+	}
+	_, actualErr := v.Value()
+	return Err[S](actualErr)
+}
+
+// Do apply f to the content of the result if the result is ok
```
samarth-aws: nit: typo: "apply" -> "applies"

cmd/argot/statistics/statistics.go
```diff
@@ -80,38 +80,36 @@ func NewFlags(args []string) (Flags, error) {
...
+	analyzer, err := loadprogram.NewState(c).Value()
 	if err != nil {
-		logGroup.Errorf("Failed to initialize state ...")
+		c.Logger.Errorf("Failed to initialize state ...")
 		return nil
 	}
-	reachableFunctions := analyzer.ReachableFunctions()
+	reachableFunctions, err := analyzer.ReachableFunctions()
+	if err != nil {
+		return fmt.Errorf("error computing reachable functions")
```
samarth-aws: nit: missing message from err variable

analysis/ptr/pointer.go
```diff
@@ -12,20 +12,69 @@
...
 	"github.com/awslabs/ar-go-tools/internal/pointer"
 	"golang.org/x/tools/go/ssa"
 	"golang.org/x/tools/go/ssa/ssautil"
 )
 
-// This file contains functions for running the pointer analysis on a program. The pointer analysis is implemented in
-// the x/tools/go/pointer package.
+// State extends the WholeProgramState with the pointer analysis information.
+type State struct {
+	*loadprogram.State
```
samarth-aws, line 34: I don't think this can modify the program state so maybe it shouldn't be a pointer

analysis/taint/taint_utils_test.go
```diff
@@ -253,40 +259,40 @@ func runTest(t *testing.T, dirName string, files []string, summarizeOnDemand boo
...
 	}
-	setupConfig(lp.Config, summarizeOnDemand)
-	state, err := dataflow.NewInitializedAnalyzerState(lp.Prog, lp.Pkgs, config.NewLogGroup(lp.Config), lp.Config)
+	resultMonad.Do(lp, func(lp *loadprogram.State) { setupConfig(lp.Config, summarizeOnDemand) })
+	ptrState := resultMonad.Bind(lp, ptr.NewState)
+	state, err := resultMonad.Bind(ptrState, dataflow.NewState).Value()
 	if err != nil {
 		t.Fatalf("failed to initialize state")
 	}
 	result, err := taint.Analyze(state)
```
samarth-aws: nit: could rename `result` here to `res` to avoid the `resultMonad` named import

samarth-aws (dismissed):

internal/funcutil/result/result.go
victornicolet: I would use `Map` for a functional operation that does the same thing, `Do` is explicitly mutating / side-effecting.

analysis/ptr/pointer.go
victornicolet: There's a lock value (atomic.Int32) in the `loadprogram.State` so it shouldn't be passed around by value

analysis/dataflow/state.go
victornicolet: There's no such instance, but this approach makes it easier to "declare" the parallel steps that need to be run in the high-level `NewState` function and then let `newFlowState` apply them in parallel. I don't think we'd gain a lot in readability by making the function calls explicit with separate go calls and increment/decrements of the waitgroup inside the `newFlowState` function. This also makes it easy to add more steps.

analysis/dataflow/state.go
victornicolet: I think one use case would be if we want to run several dataflow analyses (taint, backtrace..) on the same state, and knowing whether the `IntraProceduralPass` has to run or not. Otherwise I don't think there is a use case yet?

samarth-aws (approved):

PR #119, opened 2024-12-03
Fix annotation loading for comments in the AST

samarth-aws:
*Issue #, if available:*

*Description of changes:*
`analysisutil.VisitPackages` does not visit all the packages in the program. Instead, we need to manually parse each file in the project to get its AST.

### DIFF
--- a/analysis/loadprogram/state.go
+++ b/analysis/loadprogram/state.go
@@ -16,6 +16,11 @@ package loadprogram
 
 import (
 	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"path/filepath"
 	"sync/atomic"
 
 	"github.com/awslabs/ar-go-tools/analysis/annotations"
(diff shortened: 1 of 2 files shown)


analysis/loadprogram/state.go
```diff
@@ -173,3 +175,50 @@ func (wps *State) IncrementAndTestAlarms() bool {
...
+	var files []*ast.File
+
+	for _, dir := range dirs {
+		if err := filepath.Walk(dir, func(path string, info fs.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+
+			if info.IsDir() {
+				parsedDir, err := parser.ParseDir(fset, path, nil, parser.ParseComments)
```
victornicolet, line 194: We can use this as a hack for now, but I think we should look further into getting all the files we need, I hope we don't really have to parse files twice.

analysis/loadprogram/state.go
```diff
@@ -173,3 +175,50 @@ func (wps *State) IncrementAndTestAlarms() bool {
 func (wps *State) TestAlarmCount() bool {
 	return wps.Config.MaxAlarms <= 0 || wps.numAlarms.Load() < int32(wps.Config.MaxAlarms)
 }
+
+func allAstFiles(dirs []string, fset *token.FileSet, pkgs []*packages.Package) ([]*ast.File, error) {
+	var files []*ast.File
+
+	for _, dir := range dirs {
+		if err := filepath.Walk(dir, func(path string, info fs.FileInfo, err error) error {
```
victornicolet, line 187: nit: the closure in the condition is not very readable.

victornicolet (dismissed):

victornicolet (approved):

PR #121, opened 2024-12-06
Explicit builtin nodes

victornicolet:
This PR adds explicit nodes for calls to builtins in the dataflow graph. The effect on dataflow of builtin nodes is still implemented directly in the intraprocedural analysis, but explicit builtin nodes allows user to specify, for example, that `len` can be seen as a sanitizer.

This also extracts some ssa value marking operations out of the monotone dataflow analysis loop. Those marking operations are now moved to initialization, which should result in minor performance improvements.

### DIFF
--- a/analysis/dataflow/builtins.go
+++ b/analysis/dataflow/builtins.go
@@ -14,53 +14,111 @@
 
 package dataflow
 
-import "golang.org/x/tools/go/ssa"
+import (
+	"github.com/awslabs/ar-go-tools/internal/funcutil"
+	"golang.org/x/tools/go/ssa"
+)
 
+// handledBuiltins is the list of builtin function names that are internally handled by the analysis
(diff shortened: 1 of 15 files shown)


analysis/dataflow/builtins.go
```diff
@@ -14,126 +14,139 @@
 
 package dataflow
 
-import "golang.org/x/tools/go/ssa"
+import (
+	"github.com/awslabs/ar-go-tools/internal/funcutil"
+	"golang.org/x/tools/go/ssa"
+)
 
+// handledBuiltins is the list of builtin function names that are internally handled by the analysis
```
samarth-aws: A link to where you found these builtins would be nice so we can check the list on every major Go release

samarth-aws (dismissed):

samarth-aws (approved):

PR #122, opened 2024-12-13
Extending the IsStatic definition used by the backtrace analysis.

victornicolet:
This PR extends the function  `IsStatic` in the backtrace analysis that is used to analyze whether the endpoints of the dataflow paths are statically defined data.
This resolves a problem in which definitions like these:
```
const s1 = "foo"
const s2 = "bar"
var s = fmt.Sprintf("%s", s1, s2)
```
usage of `s1, s2` was not captured as "statically defined" because in the SSA form, `s1` and `s2` were grouped into an array first, and then a slice of the array was passed to `fmt.Sprintf`.

### DIFF
--- a/analysis/lang/values.go
+++ b/analysis/lang/values.go
@@ -171,7 +171,7 @@ func TryTupleIndexType(v types.Type, i int) types.Type {
 	return tupleType.At(i).Type()
 }
 
-// CanType checks some properties to ensure calling the Type() method on the value won't cause a sefgfault.
+// CanType checks some properties to ensure calling the Type() method on the value won't cause a segfault.
 // This seems to be a problem in the SSA.
 func CanType(v ssa.Value) (res bool) {
 	defer func() {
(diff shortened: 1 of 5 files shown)


analysis/lang/values.go
```diff
@@ -187,3 +187,99 @@ func CanType(v ssa.Value) (res bool) {
...
+// This is the case for:
+// - constants
+// - slices of constants
+// This does not analyze whether the value v is mutated; this function is useful in conjunction with the dataflow
+// analyzes for example, which would track all dataflows to the values being analyzed. For values of nodes that do
+// not have any other incoming edges, this function is sound: the dataflow analysis indirectly guarantees no data
+// is flowing from a parameter of the function, or from being written to by another function being called in the
+// function body.
+//
+// TODO: make this function usable outside of dataflow analysis
```
samarth-aws, line 202: nit: this todo isn't needed anymore

analysis/lang/values.go
```diff
@@ -187,3 +187,99 @@ func CanType(v ssa.Value) (res bool) {
...
+}
+
+// isStaticallyDefinedAlloc recognizes patterns of allocation that are static. Those are often generated during the SSA
+// translation of static var allocations in the code, where slices must be constructed by adding all elements
+// individually.
+func isStaticallyDefinedAlloc(alloc *ssa.Alloc) bool {
+	if ptrTy, isPtrTy := alloc.Type().(*types.Pointer); isPtrTy {
+		switch ptrTy.Elem().(type) {
+		case *types.Array:
+			// Check all referrers that take references and then store data are storing static data
```
samarth-aws, line 246: Just checking the referrers might not be enough since the array could be modified outside the function

analysis/lang/values.go
victornicolet: I added this TODO in this PR; this function is not meant to be checking for e.g. modifications outside of the function; it is **not** a "controlled" mutability analysis. If we want to use it outside of its current use case (values for which there is no incoming dataflow edge) then we need to do those modifications.

analysis/lang/values.go
victornicolet: Yes it's not enough in general, however if we're checking for a node in the dataflow graph that doesn't have any incoming edges, this is sufficient: any modification to the value that is being looked at outside of the function would imply a dataflow edge.

victornicolet: We could reuse your immutability analysis to make it work for any value.

samarth-aws (approved):

PR #127, opened 2025-01-08
Precondition checking analysis

victornicolet:
This PR adds a new "precondition checking analysis" (see `docs/12_syntactic.md`). This allows the user to express constraints where functions can only be called under certain conditions, and those conditions are simple expressions that depend only on the output of a function call.

Future improvements to this analysis includes:
- a proper representation for the preconditions (path formulas), both internally and in the user interface.
- some optimizations in how the path formulas are computed and checked.


analysis/config/code_identifier.go
```diff
@@ -226,12 +226,17 @@ func (cid *CodeIdentifier) MatchType(typ types.Type) bool {
 	return cid.Type == typ.String()
 }
 
-// MatchPackageAndMethod checks whether the function f matches the code identifier on the package and method fields.
+// MatchPackageAndMethod checks whether the function f matches the code identifier on the package and method fields and
+// the context of the caller.
 // It is safe to call with nil values.
-func (cid *CodeIdentifier) MatchPackageAndMethod(f *ssa.Function) bool {
+func (cid *CodeIdentifier) MatchPackageAndMethod(caller *ssa.Function, f *ssa.Function) bool {
```
samarth-aws: nit: I'd rather have a separate function called `MatchPackageAndMethodWithCaller` because we use the old version of this function so often and it would be a pain to pass nil every time

analysis/syntactic/preconditions/precond_check.go
```diff
@@ -0,0 +1,287 @@
...
+	return locs, nil
+}
+
+func analyzeFunc(state *ptr.State, spec config.CondCheckSpec, f *ssa.Function) ([]token.Position, error) {
+	state.Logger.Infof("Checking %s which calls a function matching a call identified in %s",
+		f.Name(), spec.Tag)
+	var locs []token.Position
+	for _, blk := range f.Blocks {
+		guardStrs := computeBlockGuards(blk)
+		check := sync.OnceValue(func() bool { return checkGuard(spec, guardStrs) })
```
samarth-aws: Why is the `sync.OnceValue` needed? I thought this function needs to check that the precondition is satisfied for every function call, not just once? Maybe some code comments would help

analysis/syntactic/preconditions/testdata/func-cond/main.go
samarth-aws, line 1: Might be interesting to have some tests for `goto` statements as well. Maybe also `break`s in normal and labeled loops

analysis/syntactic/preconditions/precond_check.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// computePathCond returns the conjunction of atomic propositions that holds on the specific path of
+// blocks. If statements are always the last statement in the block. The branch is deduced by looking
+// at which block is next.
+func computePathCond(blocks []*ssa.BasicBlock) []string {
+	var cond []string
+	for i := 0; i < len(blocks)-1; i++ {
+		maybeCond := blockCond(blocks[i])
+		if maybeCond.IsSome() {
```
samarth-aws, line 220: Optionals would be a neat application of this analysis. Maybe we can use it on the argot codebase itself

analysis/syntactic/preconditions/precond_check.go
```diff
@@ -0,0 +1,287 @@
...
+			if blocks[i].Succs[0].Index == blocks[i+1].Index {
```
samarth-aws, line 224: Is this block structure well-defined? We want to make sure that it won't break in the future if e.g. the ssa package changed the way blocks are ordered wrt conditional jumps. Maybe it's covered under the Go 1.0 backwards-compatibility guarantee though

analysis/syntactic/preconditions/precond_check.go
```diff
@@ -0,0 +1,287 @@
...
+	case *ssa.Extract:
+		return fmt.Sprintf("%s#%d", stringCond(value.Tuple), value.Index)
+	case *ssa.BinOp:
+		return fmt.Sprintf("%s %s %s", stringCond(value.X), value.Op, stringCond(value.Y))
+	case *ssa.UnOp:
+		return fmt.Sprintf("%s %s", value.Op, stringCond(value.X))
+	case *ssa.Const:
+		return value.String()
+	default:
+		return "<!nae!>"
```
samarth-aws, line 299: Would a panic or error value be more appropriate here to avoid silent failures?

analysis/lang/expression.go
```diff
@@ -0,0 +1,26 @@
...
+
+package lang
+
+import (
+	"golang.org/x/tools/go/ssa"
+)
+
+// An Expression can be used to manipulate expression for symbolic execution.
+// TODO: implementations
+type Expression interface {
```
samarth-aws, line 23: dead code?

samarth-aws (dismissed):

analysis/lang/expression.go
victornicolet: Dead code but placeholder for how we should improve this analysis with proper expressions instead of just lists of strings.

analysis/syntactic/preconditions/precond_check.go
victornicolet: It's expected that those will occur. For example, if I need to check that A and B are true when I call f(), it's fine if `A & B & <!nae!>` holds (at least A and B hold). The analysis also doesn't make guarantees on side-effects, it's just about what has been executed on the path and the result of some function calls, so it's ok that there are uninterpreted literals in the path formulas.

analysis/syntactic/preconditions/precond_check.go
victornicolet: That block structure is guaranteed only by the current SSA construction, but changing it would be very significant. We know that:
- if a block has a conditional, that conditional is the last statement,
- the first successor is the branch when condition is true,
- the second successor is the branch when condition is false
We have way other than looking at the successor index to determine which successor block corresponds to which branch (which is what this is testing).
Is it what you meant?

analysis/syntactic/preconditions/precond_check.go
victornicolet: It's to ensure this somewhat expensive operation is executed at most once, but maybe never. I can add comments.

analysis/syntactic/preconditions/precond_check.go
samarth-aws: Yes, that's what I meant. Thanks for clarifying!

samarth-aws: It might be worth it to put that in a comment somewhere for future reference (and to help anyone reading the code)

analysis/lang/expression.go
samarth-aws: Should it go under analysis/lang or the preconditions package? I would assume the latter unless we plan to use it in other analyses

victornicolet: I was hoping it would be a general representation for symbolic reasoning on expressions. The taint analysis actually does a little bit too and this would be used in preconditions and in the taint analysis.

samarth-aws (approved):

PR #129, opened 2025-01-09
Adding "platform" to the target specs to allow specifying a platform

victornicolet:
Targets can now optionally specify a platform. This is equivalent of adding `GOOS=platform` to built a specific target.

### DIFF
--- a/analysis/config/config.go
+++ b/analysis/config/config.go
@@ -53,7 +53,7 @@ func LoadGlobal() (*Config, error) {
 type EscapeConfig struct {
 
 	// Functions controls behavior override, keyed by .String() (e.g. command-line-arguments.main,
-	// (*package.Type).Method, etc). A value of "summarize" means process normally, "unknown" is
+	// (*package.Type).Method, etc.). A value of "summarize" means process normally, "unknown" is
 	// treat as unanalyzed, and "noop" means calls are assumed to have no escape effect (and return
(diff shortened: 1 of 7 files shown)


analysis/config/config.go
```diff
@@ -236,6 +236,8 @@ type TargetSpec struct {
 	Name string
 	// Files identifies the target's files
 	Files []string
+	// Platform identifies the target's platform
+	Platform string
```
samarth-aws, line 240: Can we support multiple platforms so that users won't have to duplicate specs?

analysis/config/validation.go
```diff
@@ -20,6 +20,9 @@ import (
 	"github.com/awslabs/ar-go-tools/internal/funcutil"
 )
 
+// A list of supported Platforms. We may want to figure out how to get it programmatically, but this should suffice for now.
```
samarth-aws: nit: link to where one can find this info would be useful

analysis/config/config.go
victornicolet: I agree, I actually started implementing with `Platform: []string` but then it became quite messy in loading the targets so for now I'd keep it 1 target = 1 platform. Only targets will need to be duplicates, the specs will refer to the various targets.

samarth-aws (approved):

PR #130, opened 2025-01-10
Adding -out flag to set output folder from cli.

victornicolet:
This PR adds a `-out` flag to the common flag. This allows users to specify the output folder for the analyses from the command line as opposed to the config file, so that different output folders can be used depending on the environment without having to change the config file.

### DIFF
--- a/analysis/config/config.go
+++ b/analysis/config/config.go
@@ -40,13 +40,21 @@ func SetGlobalConfig(filename string) {
 }
 
 // LoadGlobal loads the config file that has been set by SetGlobalConfig
-func LoadGlobal() (*Config, error) {
-	cfg, err := LoadFromFiles(configFile)
+func LoadGlobal(overrides *CmdLineOverrides) (*Config, error) {
+	cfg, err := LoadFromFiles(configFile, overrides)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load global config file %v: %v", configFile, err)
 	}
 
 	return cfg, err
+}
+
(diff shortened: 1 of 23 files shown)


analysis/config/config.go
```diff
@@ -40,15 +40,23 @@ func SetGlobalConfig(filename string) {
 }
 
 // LoadGlobal loads the config file that has been set by SetGlobalConfig
-func LoadGlobal() (*Config, error) {
-	cfg, err := LoadFromFiles(configFile)
+func LoadGlobal(overrides *CmdLineOverrides) (*Config, error) {
```
samarth-aws, line 43: nit: I don't think `CmdLineOverrides` needs to be passed as a pointer in any of these methods

samarth-aws (dismissed):

analysis/config/config.go
victornicolet: It's so that you can pass `nil` to `LoadGlobal` (and all the functions that accept overrides) to indicate nothing needs overrriding. Would you rather have an option?

samarth-aws: The empty struct also works and you wouldn't have to change much of the logic since you have the `cmdLineOverride.ReportsDir != ""` checks. Not a huge deal though, I just personally find empty structs to be more readable and safer than nil

victornicolet: Yes, I guess in this case the default values happen to be values to ignore, but if we add a parameter whose default means something (or decide that `logelevel=0` means tool should be silent) we we would have to change it

samarth-aws (approved):

PR #132, opened 2025-01-14
Contextual logging

victornicolet:
This PR adds "context" to the log message. Context is added by `PushContext` and remove by `PopContext`.
This is used to indicate which target and which problem tag the log message corresponds to.
For example, running the analyses configured in `payload/selfcheck/config.yaml` we have outputs like:

Screenshot 2025-01-14 at 5 00 10 PM

Some included changes are preparing for the addition of an "auto" tool.


analysis/taint/taint.go
```diff
@@ -167,6 +126,52 @@ func Analyze(state *dataflow.State, reqs AnalysisReqs) (AnalysisResult, error) {
 	return AnalysisResult{State: state, Graph: *state.FlowGraph, TaintFlows: taintFlows}, err
 }
 
+func runSpec(state *dataflow.State, reqs AnalysisReqs, taintSpec config.TaintSpec, taintFlows *Flows) {
```
samarth-aws, line 129: nit: Can most of this be refactored into a function in the `config` package to reduce duplicate code?

analysis/config/logging.go
```diff
@@ -63,15 +67,64 @@ func NewLogGroup(config *Config) *LogGroup {
...
+		debug:        log.New(os.Stdout, "[DEBUG]"+" ", 0),
+		info:         log.New(os.Stdout, formatutil.Green("[INFO]")+" ", 0),
+		warn:         log.New(os.Stdout, formatutil.Yellow("[WARN]")+" ", 0),
+		err:          log.New(os.Stdout, formatutil.Red("[ERROR]")+" ", 0),
 	}
 	return l
 }
 
+// PushContext pushes a context string on the prefix of the logger
+func (l *LogGroup) PushContext(s string) {
```
samarth-aws, line 81: I'm not sure how I feel about this function mutating the logger in-place without a mutex since it's possible that things will be logged in multiple threads, leading to data races. Adding a mutex may make logging slow though. I think I'd prefer that this function return a new logger

samarth-aws:
How much will we have to change this code if/when we migrate to the `log/slog` package? https://pkg.go.dev/log/slog

analysis/config/logging.go
victornicolet: That's a good point, but the state/logger is currently designed to be very stateful and not thread-safe.
If we return a new logger, we would have to significantly change where all the analysis functions get their logger, which is currently stored in the state. In general, the state is not thread safe currently, and the analyses are executed sequentially. There are exceptions where we have parallel processing but the state is only read, with the exceptions of the errors that have a mutex. 
If we want to run analyses in parallel (which would be interesting) I think we will have to do a lot of work on thread safety of the state.

analysis/taint/taint.go
victornicolet: Hmm I don't think so, I tried to refactor backtrace and taint in the same way to achieve that but it wouldn't work very well as is. Maybe with a little more effort?

victornicolet:
> How much will we have to change this code if/when we migrate to the `log/slog` package? https://pkg.go.dev/log/slog

I haven't thought about it, but I would say the context we want (target/problem tag/entrypoint) is different from the usual context (function where the logging happens) in slog so it would also be custom handling in slog.

samarth-aws (approved):

PR #137, opened 2025-02-07
New rewriting step: (reflect.Value).Call resolution

victornicolet:
This PR adds:
- a framework for rewriting the program, with helpers managing imports and AST manipulation. 
- can be used to rewrite `(reflect.Value).Call` instance into sets of static call functions for certain targets.
For example, a target can have the additional configuration:
```
reflect-value-call-instances:
      - package: "package-using-reflect"
        method: "New" # a method that receives the type from which methods are called by reflection
```
Also adds a few improvements to the cli, including the option `-file` to the `ast` command to print a file's AST.

This feature will need more testing.


analysis/lang/constructors.go
```diff
@@ -131,3 +136,263 @@ func NewUnOp(op token.Token, x dst.Expr) *dst.UnaryExpr {
...
+		return a.newPossiblyQualifiedTypeIdent(t0.Obj())
+	case *types.Struct:
+		return a.newAstStructTypeExpr(t0)
+	default:
+		panic(fmt.Sprintf("implement NewTypeExpr for %s", t.String()))
+	}
+}
+
+// newAstStructTypeExpr returns the expression representing a struct type, or an error if it could not create that
+// expression.
```
samarth-aws: nit: I think the godoc is out of date

analysis/lang/constructors.go
```diff
@@ -131,3 +136,263 @@ func NewUnOp(op token.Token, x dst.Expr) *dst.UnaryExpr {
...
+		fields = append(fields, field)
+	}
+
+	return &ast.FieldList{
+		List: fields,
+	}
+}
+
+// MaybeRefNamedType is a named type or a pointer to a named type.
+type MaybeRefNamedType struct {
```
samarth-aws: nit: the name "Maybe" is a bit confusing here and seems to imply that it's an optional. If possible, I think it would be easier to split this into two types: one for `*types.Named` and one for `*types.Pointer` instead of having the boolean `IsRef` field

analysis/refactor/statefulrewrite/reflect_value_call_rewrite.go
```diff
@@ -0,0 +1,556 @@
...
+		for _, file := range p.Syntax {
+			astBuilder := lang.NewAstBuildManager(p, file)
+			for _, node := range file.Decls {
+				astutil.Apply(node,
+					func(c *astutil.Cursor) bool { return nodeTransform(c, pts, rspec, astBuilder) }, nil)
+			}
+			originalFileName := pts.Program.Fset.File(file.FileStart).Name()
+			// if the file has changed, write the file and add it to the overlay
+			if astBuilder.FileChanged {
+				// Remove all comments to avoid problems
```
samarth-aws, line 206: Shouldn't the `dst` library handle comments?

analysis/refactor/statefulrewrite/reflect_value_call_rewrite.go
```diff
@@ -0,0 +1,556 @@
...
+
+func nodeTransform(
+	c *astutil.Cursor,
+	pts *loadprogram.State,
+	rspec RVCRewriterResolvedSpec,
+	astBuilder *lang.AstBuildManager) bool {
+	// If the current node, c.Node(), is a call to sort.Sort (or
+	// sort.Stable or sort.IsSorted), replace it with calls to
+	// obj.Less, obj.Swap, and obj.Len, where obj is the argument
+	// that was passed to sort.
```
samarth-aws: I think this comment is out-of-date. I don't see anything specific to `sort.Sort` here

analysis/refactor/statefulrewrite/reflect_value_call_rewrite.go
```diff
@@ -0,0 +1,556 @@
...
+	if !ok {
+		return reflectValueSpec{}, false
+	}
+	// Expr is of the form [_ := <callee.X>.Call(_)]
+	if callee.Sel.Name != "Call" {
+		// The method is not "Call".
+		return reflectValueSpec{}, false
+	}
+	// Expr is of the form [_ := reflectValueIdent.Call(_)]
+	reflectValueIdent, ok := callee.X.(*ast.Ident)
```
samarth-aws, line 329: How do you know that `reflectValueIdent` is the `reflect` package? Wouldn't this match any `x.Call()` calls where `x` is any valid Go identifier?

analysis/refactor/statefulrewrite/reflect_value_call_rewrite.go
```diff
@@ -0,0 +1,556 @@
...
+		clauses = append(clauses, constructSwitchCase(builder, scope, method, ccs))
+	}
+	// Generate a switch statement
+	switchStmt := &ast.SwitchStmt{
+		Tag: &ast.Ident{Name: "opName"},
+		Body: &ast.BlockStmt{
+			List: clauses,
+		},
+	}
+	// reverse order of statements
```
samarth-aws: Outdated comment? I don't see any reversal performed here

analysis/refactor/statefulrewrite/stateful_rewriter_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	spec0 := statefulrewrite.FindImpl(ptrState, config.CodeIdentifier{Method: "Start", Package: ""})
+	if spec0.IsNone() {
+		t.Fatal("no impl found")
+	}
+	spec := spec0.Value().Compile(lp)
+
+	overlay := statefulrewrite.RewriteCallsToReflectValueCall(ptrState, spec)
+	for fileName, fileContents := range overlay {
+		t.Logf("overlay %s", fileName)
```
samarth-aws, line 59: This test only tests for compilation success. We should have more correctness checks here

internal/rewrite/util.go
```diff
@@ -295,3 +299,64 @@ func forEachPackageIncludingDependencies(pkgs []*packages.Package, fn func(*pack
...
+	case *types.Array:
+		return TypeNameForGeneration(t.Elem())
+	case *types.Slice:
+		return TypeNameForGeneration(t.Elem())
+	case *types.Map:
+		return TypeNameForGeneration(t.Elem())
+	case *types.Chan:
+		return TypeNameForGeneration(t.Elem())
+	case *types.Signature:
+		return TypeNameForGeneration(t.Params().At(0).Type())
```
samarth-aws, line 361: Why only check the first param here?

analysis/lang/constructors.go
victornicolet: Yeah I can update the name and the doc. This type is a little special and has the following invariants:
- Named != nil <=> Actual != nil 
- IsRef => Actual.Elt == Named
- !IsRef => Actual == Named
If I split it into two types I have to pass around two objects instead of one in the functions that use it.

analysis/refactor/statefulrewrite/reflect_value_call_rewrite.go
victornicolet: Yes it should, but it might make the interfacing with type information more work.

analysis/refactor/statefulrewrite/reflect_value_call_rewrite.go
victornicolet: Hmm that's true, should check the type of the identifier!

analysis/refactor/statefulrewrite/reflect_value_call_rewrite.go
victornicolet: It's in reverse order of how they should appear (switch should be after the receiver assignment) but it gets reversed when the statements are inserted in the AST. I'll make it clear what this comment means

internal/rewrite/util.go
victornicolet: It's just a heuristic to generate readable variable names. I think it would get too long too quickly if we add too much information, but we could have heuristics that shorten names and look at multiple arguments. 
Handling scopes to ensure names are fresh is done in a separate function.

samarth-aws (approved):

PR #139, opened 2025-02-12
Fix tag support in syntactic analysis.

victornicolet:
The `syntactic` tool did not support tags like the `backtrace` and the `taint` analysis tool, despite having analysis problem that can be tagged in the config file.
This adds support for tags in the tool. Problems that don't have the correct tag will be ignored during the analyses by both the `struct-init` and `cond-check` analyses.

### DIFF
--- a/analysis/syntactic/structinit/structinit.go
+++ b/analysis/syntactic/structinit/structinit.go
@@ -30,6 +30,12 @@ import (
 	"golang.org/x/tools/go/ssa"
 )
 
+// AnalysisReqs groups the options of the analysis together
+type AnalysisReqs struct {
+	// Tag is the tag of the problem to analyze
+	Tag string
+}
+
 // AnalysisResult is the result of the struct-init analysis.
 type AnalysisResult struct {
 	// InitInfos is a mapping from the named struct type to its initialization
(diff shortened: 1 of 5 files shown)


analysis/syntactic/structinit/structinit.go
```diff
@@ -77,15 +83,19 @@ type InvalidWrite struct {
...
 	fns := state.ReachableFunctions()
 	if len(fns) == 0 {
 		return AnalysisResult{}, fmt.Errorf("no functions found")
 	}
-	specs := structInitSpecs(state.Config, state.Target)
+	specs := structInitSpecs(state.Config, state.Target, reqs.Tag)
 	state.Logger.Infof("%d struct-init specs to check: %s", len(specs), strings.Join(
 		funcutil.Map(specs, func(ss config.StructInitSpec) string { return ss.Tag }), ","))
+	if len(specs) == 0 {
+		state.Logger.Infof("No specs to check given the parameters.")
```
samarth-aws: nit: the word "parameters" could be confusing. Do you mean to say "No specs to check given the configuration"?

samarth-aws (dismissed):

samarth-aws (approved):

PR #141, opened 2025-02-14
Applications of Argot to external repositories: amazon-ssm-agent

victornicolet:
This PR adds a Dockerfile, python script and workflow to check program properties of an external repository (amazon-ssm-agent) that already contains an Argot configuration file.


payload/amazon-ssm-agent-check/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Copy go mod and cache dependencies
+COPY go.mod go.sum ./
+RUN go env -w GOPROXY=direct
+RUN go mod download
+
+# Build Argot
+COPY . .
+RUN go build -o bin/argot ./cmd/argot/main.go
+RUN go install github.com/awslabs/ar-go-tools/cmd/argot
```
samarth-aws, line 18: The `go install` command only installs the version of argot that is in pkg.go.dev. Should we run tests on mainline as well?

victornicolet: I think this installs the local version of argot, it's the same command that `make install` runs. If you want to install the latest pkd.go.dev I think you need to add @latest

samarth-aws (approved):

PR #147, opened 2025-02-28
Support tracking specific args in backtrace

samarth-aws:
Adds a new code identifier format.

There's a lot of special-casing in the implementation so we need to refactor this soon - especially if we want to support other types of code identifiers in the future.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/analysis/backtrace/testdata/trace-specific/config.yaml
+++ b/analysis/backtrace/testdata/trace-specific/config.yaml
@@ -1,0 +1,25 @@
+dataflow-problems:
+  slicing:
+    - backtracepoints:
+        - target:
+            kind: "call"
+            package: "github.com/awslabs/ar-go-tools/analysis/backtrace/testdata/trace-specific"
+            method: "trackSecondArg"
+            objects:
+              - kind: "argument"
+                name: "y"
+                index: 1
+                type: "string"
+        - target:
+            kind: "call"
+            package: "github.com/awslabs/ar-go-tools/analysis/backtrace/testdata/trace-specific"
+            method: "trackSecondAndThirdArg"
+            objects:
+              - kind: "argument"
+                name: "y"
+                index: 1
+                type: "uint"
+              - kind: "argument"
+                name: "z"
+                index: 2
+                type: "string"
--- a/analysis/backtrace/testdata/trace-specific/config.yaml
+++ b/analysis/backtrace/testdata/trace-specific/config.yaml
@@ -23,3 +23,12 @@ dataflow-problems:
                 name: "z"
                 index: 2
                 type: "string"
+        - target:
+            kind: "call"
+            package: "github.com/awslabs/ar-go-tools/analysis/backtrace/testdata/trace-specific"
+            method: "trackOnlyVariadic"
+            objects:
+              - kind: "argument"
+                name: "ys"
+                index: 1
+                type: "...string"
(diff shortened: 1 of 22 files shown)


analysis/config/code_identifier.go
```diff
@@ -173,6 +296,68 @@ func (cid *CodeIdentifier) equalOnNonEmptyFields(cidRef CodeIdentifier) bool {
...
+//
+//gocyclo:ignore
+func (cid *CodeIdentifier) equalNewFormat(cidRef CodeIdentifier) bool {
+	// If cidRef is in the new format, cid will also be in the new format
+	if cidRef.Target.Kind != cid.Target.Kind {
+		return false
+	}
+	switch cidRef.Target.Kind {
+	case CallKind:
+		// package and method must match
```
victornicolet, line 310: The old format was testing equality on non-empty fields, do we assume here that every field is fully specified? What if I don't want to specify the package?

analysis/config/code_identifier.go
```diff
@@ -149,6 +268,10 @@ func compileRegexes(cid CodeIdentifier) CodeIdentifier {
 //
 //gocyclo:ignore
 func (cid *CodeIdentifier) equalOnNonEmptyFields(cidRef CodeIdentifier) bool {
+	if isNewFormat(cidRef) {
```
victornicolet, line 271: Here only the `cidRef` is in new format, not necessarily the `cid`, but the `cid.equalNewFormat` seems to assume both are in the new format?

analysis/config/code_identifier.go
samarth-aws: Yes, we assume that every field is fully specified. I think this will reduce the possibility for user error.

victornicolet: Ok I think then we should (not now but later) also allow regexes for at least package, we've been commonly using things like `method: (Infof|Debugf)` for any logging method. This would allow user to specify any package, and because it's explicit reduce the possibility of user error.

analysis/syntactic/structinit/structinit.go
```diff
@@ -230,11 +230,12 @@ func initInfos(
 	allocs []alloced,
 	specs []config.StructInitSpec) (map[*types.Named]InitInfo, error) {
 	infos := make(map[*types.Named]InitInfo)
-	initialized := make(map[config.CodeIdentifier]bool)
+	initialized := make(map[structId]bool)
```
victornicolet, line 233: Side note: I think it's neat to use code identifiers wherever we need a flexible way to identify code. I think code identifiers should support clearly identifying a struct later on.

analysis/backtrace/testdata/trace-specific/config.yaml
```diff
@@ -0,0 +1,25 @@
...
+                type: "string"
+        - target:
+            kind: "call"
+            package: "github.com/awslabs/ar-go-tools/analysis/backtrace/testdata/trace-specific"
+            method: "trackSecondAndThirdArg"
+            objects:
+              - kind: "argument"
+                name: "y"
+                index: 1
+                type: "uint"
```
victornicolet, line 21: There should be a test for variadic arguments to make sure this config format is flexible enough to capture them.

analysis/config/code_identifier.go
samarth-aws: True, I'll add a check for `cid` as well

analysis/config/code_identifier.go
```diff
@@ -149,6 +268,16 @@ func compileRegexes(cid CodeIdentifier) CodeIdentifier {
...
+		if cid.Target.Method == "trackOnlyVariadic" && cid.Target.Objects[0].Index == 1 && cidRef.Target.Method == "trackOnlyVariadic" && cidRef.Target.Objects[0].Index == 1 {
+			if !cid.equalNewFormat(cidRef) {
+				panic("test")
```
victornicolet: Is that a leftover panic from testing?

samarth-aws: Oops, yes it is

victornicolet (approved):

PR #150, opened 2025-03-26
Implementing must-reinit check

victornicolet:
This PR adds functionality to the `struct-init` syntactic check problem. Now, users can add a `must-reinit` specification that instructs Argot to check the results of some specified function call has its field reinitialized. The fields that must be reinitialized are the same fields that are constrained in the `struct-init` problem the `must-reinit` specification belongs to.

See doc/12_syntactic.md for an example and more information.

### DIFF
--- a/analysis/lang/instructions.go
+++ b/analysis/lang/instructions.go
@@ -151,10 +151,10 @@ func InstrSwitch(visitor InstrOp, instr ssa.Instruction) {
 
 // Utilities for working with blocks and instructions
 
-// ParentBlockIndex returns the parent block of an instruction with the index of the instruction in it.
+// IndexInEnclosingBlock returns the parent block of an instruction with the index of the instruction in it.
 //
 //	If the instruction is nil, or the parent function is nil, then it returns nil.
-func ParentBlockIndex(instr ssa.Instruction) (*ssa.BasicBlock, int) {
+func IndexInEnclosingBlock(instr ssa.Instruction) (*ssa.BasicBlock, int) {
 	if instr == nil || instr.Block() == nil {
 		return nil, 0
 	}
--- a/analysis/lang/instructions.go
+++ b/analysis/lang/instructions.go
@@ -151,6 +151,23 @@ func InstrSwitch(visitor InstrOp, instr ssa.Instruction) {
 
 // Utilities for working with blocks and instructions
 
+// ParentBlockIndex returns the parent block of an instruction with the index of the instruction in it.
+//
+//	If the instruction is nil, or the parent function is nil, then it returns nil.
+func ParentBlockIndex(instr ssa.Instruction) (*ssa.BasicBlock, int) {
+	if instr == nil || instr.Block() == nil {
+		return nil, 0
+	}
+	block := instr.Block()
+	for i, blockInstr := range block.Instrs {
+		if blockInstr == instr {
+			return block, i
+		}
+	}
+	// Not found in own parent?
+	panic("internal SSA representation broken.")
+}
+
 // LastInstr returns the last instruction in a block. There is always a last instruction for a reachable block.
 // Returns nil for an empty block (a block can be empty if it is non-reachable)
 func LastInstr(block *ssa.BasicBlock) ssa.Instruction {
--- a/analysis/syntactic/structinit/structinit.go
+++ b/analysis/syntactic/structinit/structinit.go
@@ -25,6 +25,7 @@ import (
 	"github.com/awslabs/ar-go-tools/analysis/lang"
 	"github.com/awslabs/ar-go-tools/analysis/ptr"
 	"github.com/awslabs/ar-go-tools/analysis/summaries"
+	"github.com/awslabs/ar-go-tools/internal/analysisutil"
 	"github.com/awslabs/ar-go-tools/internal/formatutil"
 	"github.com/awslabs/ar-go-tools/internal/funcutil"
 	"golang.org/x/tools/go/ssa"
@@ -52,6 +53,8 @@ type InitInfo struct {
 	// InvalidWrites is a mapping of the struct field to all the invalid writes
 	// to that field.
 	InvalidWrites map[*types.Var][]InvalidWrite
+	// BadReinits is a list of bad reinitializations
+	BadReinits []BadReinit
 	// fieldExpectedValue is a mapping of the struct field to the concrete value it
 	// should be initialized to according to the spec.
 	//
@@ -78,6 +81,15 @@ type InvalidWrite struct {
 	Want ssa.Value
 	// Instr is the instruction performing the write.
 	Instr ssa.Instruction
+	// Pos is the position of the instruction.
+	Pos token.Position
+}
+
+// BadReinit is a function call resulting in a struct that should have specific
+// fields reinitialized, but it hasn't.
+type BadReinit struct {
+	// Call is the function call instruction.
+	Call ssa.Instruction
 	// Pos is the position of the instruction.
 	Pos token.Position
 }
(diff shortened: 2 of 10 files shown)


analysis/syntactic/structinit/structinit.go
```diff
@@ -822,3 +853,139 @@ func isFiltered(spec config.StructInitSpec, f *ssa.Function) bool {
...
+			}
+		})
+	}
+	return badReinits
+}
+
+// checkStructTyp extracts the types.Named type of a struct type or a pointer to a struct type.
+// This is for checking reintiializations: we only check them for function that returns the proper
+// named struct type.
+func checkStructTyp(typ types.Type, structToNamed map[*types.Struct]*types.Named) *types.Named {
```
samarth-aws: nit: This function only extracts the name of the struct type and doesn't really check anything yet. The name should probably be `namedStructTyp`

analysis/lang/instructions.go
```diff
@@ -151,6 +151,23 @@ func InstrSwitch(visitor InstrOp, instr ssa.Instruction) {
 
 // Utilities for working with blocks and instructions
 
+// ParentBlockIndex returns the parent block of an instruction with the index of the instruction in it.
+//
+//	If the instruction is nil, or the parent function is nil, then it returns nil.
+func ParentBlockIndex(instr ssa.Instruction) (*ssa.BasicBlock, int) {
```
samarth-aws: nit: `ParentBlockIndex` is a bit confusing to me because I thought "Parent" referred to `instr.Parent()` which is a function, not the "parent block" which is the block to which the instruction belongs. I think renaming this function to `EnclosingBlockIndex` or something similar will be more clear

analysis/syntactic/structinit/structinit.go
```diff
@@ -822,3 +853,139 @@ func isFiltered(spec config.StructInitSpec, f *ssa.Function) bool {
...
+			for _, callee := range callees {
+				if callSpec.MatchPackageAndMethod(callee.Callee) {
+					mustCheckFor = append(mustCheckFor, spec)
+				}
+			}
+		}
+	}
+
+	if len(mustCheckFor) == 0 {
+		return funcutil.None[BadReinit]() /* Nothing to do */
```
samarth-aws, line 931: Should there be an error/warning if there is a `must-reinit` spec but no function matched?

analysis/syntactic/structinit/structinit.go
```diff
@@ -822,3 +853,139 @@ func isFiltered(spec config.StructInitSpec, f *ssa.Function) bool {
...
+	}
+	// For each spec, the statements following directly the call MUST write the fields
```
samarth-aws, line 933: It's important to mention that we only check the statements following directly after the call in the *same basic block* as the call

analysis/syntactic/structinit/structinit.go
```diff
@@ -822,3 +853,139 @@ func isFiltered(spec config.StructInitSpec, f *ssa.Function) bool {
...
+	// that are specified in the spec.
+	block, index := lang.ParentBlockIndex(callInstr)
+	fieldsToReinit := map[string]bool{}
+	for _, spec := range mustCheckFor {
+		for _, fieldSpec := range spec.FieldsSet {
+			fieldsToReinit[fieldSpec.Field] = true
+		}
+	}
+	var callVal ssa.Value
+	callVal = callInstr
```
samarth-aws, line 944: nit: `callVal` to me would refer to the value of the call instruction itself (like an `*ssa.Function`), but this is really the return value of the call (I know these things are conflated in the SSA library but to me they're distinct). I think a name like like `callRes` would make this clearer. You can also simplify this to the one-liner `callRes := callInstr.(ssa.Value)`

analysis/lang/instructions.go
```diff
@@ -151,6 +151,23 @@ func InstrSwitch(visitor InstrOp, instr ssa.Instruction) {
...
+	if instr == nil || instr.Block() == nil {
+		return nil, 0
```
samarth-aws, line 159: Should this case be an error? When would we ever pass a `nil` instruction to this function? We're not handling the nil case on the caller side either

analysis/syntactic/structinit/structinit.go
```diff
@@ -822,3 +853,139 @@ func isFiltered(spec config.StructInitSpec, f *ssa.Function) bool {
...
+					if _, ok := badReinits[namedType]; !ok {
+						badReinits[namedType] = []BadReinit{}
+					}
+					badReinits[namedType] = append(badReinits[namedType],
+						maybeBadReinit.Value())
+				}
+			}
+		})
+	}
+	return badReinits
```
samarth-aws, line 893: Should there be an error if there is a `must-reinit` spec but `len(badReinits) == 0`? The analysis would silently succeed in the trivial case where `len(fns) == 0`

doc/12_syntactic.md
```diff
@@ -40,6 +40,41 @@ initialization information for crypto/tls.Config:
...
+config := dependency.generateTlsConfig()
+config.MinVersion = tls.VersionTLS12
+```
+but this will not pass:
+```go
+config := dependency.generateTlsConfig()
+handleConfig(config)
+config.MinVersion = tls.VersionTLS12
+```
+because the fields MUST be reinitialized in the same block as the call to the function, and immediately after. If any other instruction precedes the field assignment, Argot will generate a finding. In particular, assigning other fields should be done after having assigned the fields that are tracked by the `struct-init` problem for which `must-reinit` is specified.
```
samarth-aws, line 77: We should say that this is more of a "linting" check since it is purely syntactic and can be unsound due to global variables, parallelism, etc.

analysis/syntactic/structinit/structinit.go
victornicolet: This is also true for all the analyses, they succeed when there is nothing to analyze and I it should  not return an error. Some analyses are run as "sanity checks" just in case someone introduced something that needs analyzing. 
Sanity checking that the analyses are running should be a different mechanism (e.g. introduce changes that should trigger alarms).

analysis/syntactic/structinit/structinit.go
victornicolet: The simplification `callRes := callInstr.(ssa.Value)` is semantically different and in this case doesn't work (`callInstr.(ssa.Value)` is not a valid value, callInstr is not of interface type). `callVal` in this case is kind of the call result value directly, but can also follow a store into a local value, so I didn't want to call it exactly `callRes`.

analysis/syntactic/structinit/structinit.go
victornicolet: That's true. 
Although technically if the statements following directly after the call MUST be writes to the field, they can't be statements that generate a new block, so they have to be in the same block.

analysis/syntactic/structinit/structinit.go
victornicolet: Not really, this check is run for every possible call in the program that returns the struct type of interest, but it's not necessarily a `must-reinit` call. There will be many calls where this is the expected path.

analysis/lang/instructions.go
victornicolet: From a safe programming point of view this is where I want to ensure the instr is safe to avoid segfaults in the code following. An instruction that has those things nil probably never occurs, but I don't want my code to segfault.

doc/12_syntactic.md
victornicolet: Should that wording go for all the syntactic checks? We're not claiming things about program behavior, we're claiming we're running syntactic checks?

samarth-aws: True, we should put a disclaimer at the top of that file

samarth-aws (approved):

PR #153, opened 2025-04-03
Track incomplete struct initializations in structinit

samarth-aws:
*Issue #, if available:*

*Description of changes:*
The structinit analysis now reports incomplete initializations of a struct, where only some fields are initialized, instead of only reporting empty initializations of a struct (previously called "zero-alloc"). The fields we track in the analysis are only the fields specified in the config.


internal/analysisutil/analysisutil.go
```diff
@@ -110,35 +110,48 @@ func FindValuePackage(n ssa.Value) fn.Optional[string] {
...
-func FieldAddrFieldInfo(fieldAddr *ssa.FieldAddr) (string, bool) {
+func FieldAddrFieldInfo(fieldAddr *ssa.FieldAddr) (FieldInfo, bool) {
 	return GetFieldInfoFromType(fieldAddr.X.Type().Underlying(), fieldAddr.Field)
 }
 
 // FieldFieldInfo finds the name of a field access in ssa.Field
 // if it cannot find a proper field name, returns "?".
 // The boolean indicates whether this field is embedded or not.
-func FieldFieldInfo(fieldAddr *ssa.Field) (string, bool) {
+func FieldFieldInfo(fieldAddr *ssa.Field) (FieldInfo, bool) {
```
victornicolet: Can this ever return false for the boolean part? If it cannot, which was initially the design, I would return only `FieldInfo`. Having to handle the `ok` part creates some verbose code.

analysis/config/code_identifier.go
```diff
@@ -391,6 +393,9 @@ func (cid *CodeIdentifier) MatchType(typ types.Type) bool {
 		return cid.Type == ""
 	}
 	if named, ok := typ.(*types.Named); ok {
+		if named == nil {
```
victornicolet: nit: should have a condition on `typ == nil` return false earlier?

analysis/syntactic/structinit/struct-types.go
```diff
@@ -0,0 +1,139 @@
...
+	if t == nil {
+		return structType{}, false
+	}
+	if t.Underlying() == nil {
+		return structType{}, false
+	}
+
+	typ := t
+	isPtr := false
+	if ptr, ok := t.Underlying().(*types.Pointer); ok {
```
victornicolet: We use this quite often, may having a `typ, isPtr := lang.TryDerefTyp` function would be useful for reading

analysis/syntactic/structinit/struct-types.go
```diff
@@ -0,0 +1,139 @@
...
+// (e.g., "<pkg-path>/structinit.structType") and its
+// underlying struct type (e.g. "struct { strct: [...] }").
+//
+// named can be nil if the struct does not have a named type
+// (i.e., it is anonymous).
+type structType struct {
+	strct   *types.Struct
+	named   *types.Named
+	isPtr   bool
+	isField bool
```
victornicolet, line 35: nit: it's a little confusing to have `structType` and `types.Struct` and there is a `isField` in `structType` (what does that mean?). Not a big deal since it's private

analysis/syntactic/structinit/struct-types.go
```diff
@@ -0,0 +1,139 @@
...
+
+// isStructType returns the structType of t and true if t is a struct type, otherwise false.
+func isStructType(t types.Type) (structType, bool) {
+	return isStructTypeHelper(t, false)
+}
+
+// isStructFieldType returns the structType of struct field type t and true if t is a struct type,
+// otherwise false.
+func isStructFieldType(t types.Type) (structType, bool) {
+	return isStructTypeHelper(t, true)
```
victornicolet, line 58: I don't understand the logic of this triple of functions where the caller needs to pick a function in order to just put a boolean value in the struct, instead of letting the caller set the value `isField`?

victornicolet:
Here is a simple test case that returns a false positive:
```
type target struct {
	x int
}

type nestedTarget struct {
	t target
	x int
}

type doubleNestedPtrStruct struct {
	t       *nestedTarget
	surname string
}

func testMultiLevelNesting() {
	ex13 := doubleNestedPtrStruct{} // false positive here!
	fmt.Println(ex13.surname)
	targetConsume(ex13.t.t) // ex13.t is nil, ex13.t.t cannot have been allocated
}
```

internal/analysisutil/analysisutil.go
samarth-aws: Oh good point, I'm not sure why I kept the boolean

analysis/syntactic/structinit/struct-types.go
samarth-aws: This type makes it more convenient to check if a struct value has a named type.
`isField` indicates that the type is of a value that is a field of a struct which is handled specially in some cases. I agree that including this information when referring to types is confusing - I'll think about a better way to represent the information.

analysis/syntactic/structinit/struct-types.go
samarth-aws: I used to have a boolean flag for `isField` but I think this is more readable because it's difficult to tell what `isStructType(t, true)` means whereas `isStructFieldType(t)` is more clear in my opinion

analysis/config/code_identifier.go
samarth-aws: This is an annoying part of Go where `typ` is not nil because it has an underlying type of `*types.Named` that is nil, hence the need for the extra check. I'll add a comment.

analysis/syntactic/structinit/struct-types.go
samarth-aws: I agree but I can address this in a future PR since it's used by multiple analyses

victornicolet (dismissed):

victornicolet:
It seems the filters for a struct-init problem are not used to filter out the incomplete struct initializations.

victornicolet (approved):

PR #157, opened 2025-04-15
Changes to CLI to enable faster loading

victornicolet:
This PR makes major changes to how the cli works internally, although the interface is mostly unchanged. 
Instead of building the whole program, running the pointer analysis and building basic dataflow analysis information, the cli now loads only with the config.
The cli user can build the parts they need later:
- to build the whole program, run `> load-program`. 
- to run the pointer analysis, run `> run-pointer`
- to run the dataflow analysis, run `> run-dataflow` (this doesn't run the intra-procedural analyses which are still run on demand). 
Additionally, a command allows you to load individual packages: `> load-pkg ./analysis/...` loads all packages in `analysis`, but doesn't build the SSA. More commands will be added to allow inspecting types and syntax in packages without SSA.

Internally, the `session` struct now carries the information that is being updated by the command instead of `*state`. Commands inspect the session information to decide whether additional analyses need to run (e.g. dataflow also runs pointer analysis if the information is not present). 

This PR will be followed with another PR updating the documentation where necessary and adding commands to inspect packages that have not been loaded with SSA.


cmd/argot/cli/package.go
```diff
@@ -0,0 +1,63 @@
...
+			writeFmt(tt, "Files:\n\t%s\n", strings.Join(pkg.GoFiles, "\n\t- "))
+		}
+		if printimports {
+			writeFmt(tt, "Imports:\n\t%s\n", strings.Join(maps.Keys(pkg.Imports), "\n\t"))
+		}
+	}
+	return false
+}
+
+func findPkg(sess *session, name string) []*packages.Package {
```
samarth-aws: This is a bit confusing - so based on the type signature there can be multiple packages with the same name? Would it be more useful to use the package path instead?

internal/funcutil/result/result.go
```diff
@@ -50,6 +54,7 @@ func (s err[T]) ValueOr(defaultVal T) T { return defaultVal }
 func (s err[T]) Value() (*T, error)     { return nil, s.value }
 func (s err[T]) IsOk() bool             { return false }
 func (s err[T]) IsErr() bool            { return true }
+func (s err[T]) Unwrap() *T             { panic(s.value) }
```
samarth-aws: nit: this could be a nicer error message like "fatal error: tried to unwrap error value %v"

cmd/argot/cli/state.go
```diff
@@ -15,73 +15,364 @@
...
-type serverState struct {
-	// the args (the path to the program to load
-	Args []string
-
-	InitialPackages []*packages.Package
+// session stores state information about the current cli session
+type session struct {
+	originalFlags tools.CommonFlags
 
+	Args       []string
```
samarth-aws: nit: It doesn't make sense to have public fields in a private struct

cmd/argot/cli/state.go
```diff
@@ -15,73 +15,364 @@
...
+	if s.LPState == nil {
+		return nil, fmt.Errorf("listing reachable functions requires at least a loaded program")
+	}
+	if s.DFState != nil {
+		return s.DFState.ReachableFunctions(), nil
+	}
+	if s.PtrState != nil {
+		return s.PtrState.ReachableFunctions(), nil
+	}
+	return s.LPState.ReachableFunctions()
```
samarth-aws: Doesn't have to be in this PR but maybe there's a nicer interface we can use for patterns like this where we want the "best" result from a state based on what state we've computed so far. Ideally we can just call `state.ReachableFunctions()`.

cmd/argot/cli/state.go
```diff
@@ -15,73 +15,364 @@
...
+			return pConfig, false, nil
+		}
+		err = s.attemptSettingConfig(&pConfig, dir, "config.json")
+	}
+	return pConfig, false, err
+}
+
+func (s *session) attemptSettingConfig(pConfig **config.Config, dir string, filename string) error {
+	configFile := filepath.Join(dir, filename)
+	config.SetGlobalConfig(configFile)
```
samarth-aws, line 145: nit: why set the global config instead of just loading it from a file?

cmd/argot/cli/state.go
```diff
@@ -15,73 +15,364 @@
...
-	writeFmt(tt, "flow graph built? : %t\n", c.FlowGraph.IsBuilt())
+	if s.CurrentFunction != nil {
+		fName = s.CurrentFunction.String()
+	}
+	writeFmt(tt, "Program path          : %s\n", strings.Join(s.Args, " "))
+	writeFmt(tt, "Config path           : %s\n", s.ConfigPath)
+	writeFmt(tt, "Working dir           : %s\n", wd)
+	writeFmt(tt, "Focused function      : %s\n", fName)
+	if s.LPState != nil {
+		writeFmt(tt, "┌────────── SSA ──────")
```
samarth-aws, line 340: I like this output! I wonder if we can do a similar thing for our other analysis tools? Computing the state should ideally look the same across Argot

cmd/argot/cli/analysis.go
```diff
@@ -288,21 +323,27 @@ func cmdAst(tt *term.Terminal, c *dataflow.State, command Command, withTest bool
 		return false
 	}
 
+	if !sess.hasProgram() {
+		WriteErr(tt, "Need a program to show the AST for.")
+		WriteErr(tt, "You should at least loadprogam")
+		cmdAst(tt, nil, command, withTest)
```
samarth-aws, line 329: Why not return here? Won't the CLI panic if there's no program?

victornicolet: A program is a whole program; now we can also load individual packages. This function will need to be updated to show AST using packages too if there are some packages available.

cmd/argot/cli/state.go
victornicolet: Some old stuff I didn't want to touch yet for fear of breaking something. I think it should be safe with the new `session` type that is more functional.

cmd/argot/cli/state.go
victornicolet: Yes that's true. In this case, reachable functions needs at least a whole program. With the introduction of package-only loading we'll have to be mindful about what a command will compute.

cmd/argot/cli/state.go
victornicolet: Yes in this case you're right but exposing public fields in private struct can be useful (e.g. you have public functions returning a struct of private type, but let the caller have access to the fields).

samarth-aws (approved):

PR #160, opened 2025-06-10
fix typo & wrong function signatures in standard_library.go

zcysky:
*Issue #, if available:* N/A 

*Description of changes:*  Update the `standard_library.go` file, for correcting the mistakes in function signatures:
1. Typo: 
- flat.DurationVar → flag.DurationVar (Incorrect package name)
- runtime.FuncForPc → runtime.FuncForPC (Incorrect casing)
- (*archive/tar.Writer).Writer → (*archive/tar.Writer).Write (Incorrect method name)
- (reflect.Value).SetMapIndex(key, elem Value) → (*reflect.Value).SetMapIndex (Arguments removed from name)
- crypto/sha256.Sum244 → crypto/sha256.Sum224

2. Correction of methods:
- time.Before → (time.Time).Before (Correctly identified as a method on the Time type)
- (encoding/base64).Strict → (encoding/base64.Encoding).Strict (Specified the correct receiver type, Encoding)

### DIFF
--- a/analysis/summaries/standard_library.go
+++ b/analysis/summaries/standard_library.go
@@ -1631,7 +1631,7 @@ var summaryText = map[string]Summarizer{}
 
 var summaryTime = map[string]Summarizer{
 	"time.After":  SingleVarArgPropagation,
-	"(time.Time).Before": SingleVarArgPropagation,
+	"time.Before": SingleVarArgPropagation,
 	// func Parse(layout, value string) (Time, error)
 	"time.Parse": Summary{
 		[][]int{{0}, {1}},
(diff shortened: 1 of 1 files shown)


analysis/summaries/standard_library.go
```diff
@@ -1631,7 +1631,7 @@ var summaryText = map[string]Summarizer{}
 
 var summaryTime = map[string]Summarizer{
 	"time.After":  SingleVarArgPropagation,
-	"(time.Time).Before": SingleVarArgPropagation,
```
victornicolet: So the other (time.Time).Before was correct?

zcysky: Yes, I ignored there's another definition here, so that one is good.

victornicolet (approved):

samarth-aws (approved):

PR #163, opened 2025-07-28
add timing for intra-procedural

zcysky:
*Issue #, if available:* N/A

*Description of changes:*

Add the time-out check for `runIntraProcedural` in `inter_procedural.go` and `intra_procedural.go`. When it times-out, it will fall back to `BuildFullGraph`, where every input param can flow to everywhere of the function.


analysis/dataflow/intra_procedural.go
```diff
@@ -80,17 +81,239 @@ func IntraProceduralAnalysis(state *State,
...
+		sm.Constructed = true
+		elapsed := time.Since(start)
+
+		return elapsed, nil
+	}
+}
+
+// // runOriginalAnalysis contains the original RunIntraProcedural logic
+// it is temporarily removed by the timing feature
+// func runOriginalAnalysis(a *State, sm *SummaryGraph) error {
```
victornicolet, line 255: Instead of commenting out this function in the code and implementing a copy, can we add the context to the signature? Does the cancellation checking have a significant impact on performance? 
If not, we can just add the checks in the original function and keep the original function.
A good name for this function could be `cancellableIntraproceduralAnalysis`

analysis/dataflow/intra_procedural.go
```diff
@@ -109,24 +332,64 @@ func RunIntraProcedural(a *State, sm *SummaryGraph) (time.Duration, error) {
...
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
+
 	// Once the state is initialized, we call the forward iterative monotone framework analysis. The algorithm is
 	// defined generally in the lang package, but all the details, including transfer functions, are in the
 	// single_function_monotone_analysis.go file
+	// This is the most time-consuming part of the analysis
 	lang.RunForwardIterative(state, sm.Parent)
```
victornicolet, line 371: In the future, you could also pass the context to the forward iterative part so that it can also stop when getting a cancel message.

analysis/dataflow/intra_procedural.go
```diff
@@ -109,24 +332,64 @@ func RunIntraProcedural(a *State, sm *SummaryGraph) (time.Duration, error) {
 		postBlockCallback:   sm.postBlockCallBack,
 	}
 
+	// Check for cancellation after initialization
```
victornicolet, line 335: Nit: wrap cancellation logic in a small function and you can write this as:
```
if e := checkCancel(ctx); e != nil {
 return e;
}
```

analysis/dataflow/intra_procedural.go
```diff
@@ -138,9 +401,9 @@ func RunIntraProcedural(a *State, sm *SummaryGraph) (time.Duration, error) {
 	// If we have errors, return one. This is sufficient to warn the user that the results are incorrect.
 	// TODO: manage error messages for better debugging
 	for _, err := range state.errors {
-		return time.Since(start), fmt.Errorf("error in intraprocedural analysis: %w", err)
+		return fmt.Errorf("error in intraprocedural analysis: %w", err)
```
victornicolet, line 404: Why not return the time when it completes?

analysis/dataflow/timeout_test.go
```diff
@@ -0,0 +1,275 @@
...
+	"testing"
+	"time"
+
+	"github.com/awslabs/ar-go-tools/analysis/dataflow"
+	"github.com/awslabs/ar-go-tools/analysis/ptr"
+	"github.com/awslabs/ar-go-tools/internal/analysistest"
+	"github.com/awslabs/ar-go-tools/internal/funcutil/result"
+	"golang.org/x/tools/go/ssa"
+)
+
```
victornicolet, line 29: In this file, we should avoid running the tests all the time, have a look at https://pkg.go.dev/testing#hdr-Skipping on how to skip tests.

analysis/dataflow/intra_procedural.go
zcysky: There is a `runOriginalAnalysisWithContext` afterwards. Agree with renaming the function.

PR #164, opened 2025-07-31
Check external summary soundness

zcysky:
*Issue #, if available:* N/A

*Description of changes:* After loading the external summaries, the `checkSummarySoundness` will use 2 intra-procedural analysis to check its soundness and make a log.

In future, this can be replaced with a decision procedural among a set of candidate analyses. 


analysis/dataflow/function_summary_graph_nodes.go
```diff
@@ -630,13 +630,14 @@ type ReturnValNode struct {
 	index  int // when a function returns a tuple, a return node represents a single indexed Value
 	parent *SummaryGraph
 	in     map[GraphNode]EdgeInfo
+	out    map[GraphNode][]EdgeInfo
```
samarth-aws, line 633: Return nodes shouldn't have outgoing intra-procedural edges. We only compute dataflow from return nodes in the inter-procedural step

analysis/dataflow/inter_procedural.go
```diff
@@ -154,6 +155,29 @@ func (g *InterProceduralFlowGraph) BuildGraph() {
...
+								formatutil.SanitizeRepr(node.Callee()), reason)
+							// Continue with the summary, but mark it as potentially problematic
+							// In the future, we might want to create a new summary or take other actions
+						} else {
+							logger.Debugf("External summary for %s is sound: %s",
+								formatutil.SanitizeRepr(node.Callee()), reason)
+
+							// If we need to dive deeper into callees, log this information
+							if len(needsDeeperCheck) > 0 {
+								logger.Debugf("Summary for %s requires deeper analysis of %d callees",
```
samarth-aws, line 172: Shouldn't the soundness check automatically dive deeper into callees? This is part of the interprocedural analysis after all

analysis/dataflow/inter_procedural.go
```diff
@@ -338,11 +362,44 @@ func (g *InterProceduralFlowGraph) RunVisitorOnEntryPoints(visitor Visitor, spec
...
-		visitor.Visit(g.AnalyzerState, entry)
+
+		// Run visitor with timeout protection
+		g.runVisitorWithTimeout(visitor, entry, i+1, len(entryPoints))
 		i++
 	}
 }
 
+// runVisitorWithTimeout runs the visitor on an entry point with timeout protection
+func (g *InterProceduralFlowGraph) runVisitorWithTimeout(visitor Visitor, entry NodeWithTrace, current, total int) {
```
samarth-aws, line 373: What are `current` and `total` used for?

analysis/dataflow/inter_procedural.go
```diff
@@ -486,6 +543,8 @@ func (g *InterProceduralFlowGraph) resolveCalleeSummary(node *CallNode,
 	if calleeSummary != nil && !calleeSummary.Constructed {
 		if shortSummary, isPredefined := summaries.SummaryOfFunc(node.Callee()); isPredefined {
 			calleeSummary.PopulateGraphFromSummary(shortSummary, false)
+			// Mark predefined summaries as sound
+			calleeSummary.IsSound = true
```
samarth-aws, line 547: This only holds if we check beforehand that our predefined summaries are sound. We should add this check to the CI later

analysis/dataflow/inter_procedural.go
```diff
@@ -650,6 +709,1040 @@ func UnwindCallStackToFunc(stack *CallStack, f *ssa.Function) *CallStack {
...
+	g.AnalyzerState.Logger.Debugf("Sp == Sg: %v", g.compareSummaries(Sp, Sg))
+	g.AnalyzerState.Logger.Debugf("Sp ⊆ Su: %v", g.isSummarySubset(Sp, Su))
+	g.AnalyzerState.Logger.Debugf("Su ⊆ Sg: %v", g.isSummarySubset(Su, Sg))
+	g.AnalyzerState.Logger.Debugf("Su == Sg: %v", g.compareSummaries(Su, Sg))
+	g.AnalyzerState.Logger.Debugf("Sp == Su: %v", g.compareSummaries(Sp, Su))
+
+	g.AnalyzerState.Logger.Debugf("=== END SUMMARY FLOWS DEBUG ===")
+}
+
+// // logSummaryFlows logs detailed flow information for a single summary
```
samarth-aws, line 739: nit: you should delete commented out code

analysis/dataflow/inter_procedural.go
```diff
@@ -650,6 +709,1040 @@ func UnwindCallStackToFunc(stack *CallStack, f *ssa.Function) *CallStack {
...
+		return true, "Summary is sound: Sp = Sg, these are the targeted dataflow facts", nil
+	} else if spSubsetOfSu && suSubsetOfSg && !suEqualsSg {
+		// Case 2: Sp ⊆ Su ⊂ Sg
+		// We need to check evidence by diving deeper into callees
+		for _, calleeNodes := range Su.Callees {
+			for _, callNode := range calleeNodes {
+				if callNode.Callee() != nil {
+					// For each callee Gi, we need to check (Su∖Sp)∩Gi
+					// Add this callee to the map of functions to analyze deeper
+					calleesDiveDeeperMap[callNode.Callee()] = g.createIntersectionSummary(Su, Sp, callNode.Callee())
```
samarth-aws, line 865: Why not implement this logic in this function? Returning a map here seems like a confusing API because how is the caller supposed to know what to do with it?

analysis/dataflow/inter_procedural.go
```diff
@@ -650,6 +709,1040 @@ func UnwindCallStackToFunc(stack *CallStack, f *ssa.Function) *CallStack {
...
+		}
+	}
+
+	// Validate summary state after creation/cloning
+	if summary.Parent == nil {
+		panic(fmt.Sprintf("summary.Parent is nil for function %v", function))
+	}
+
+	// Check if function has valid implementation
+	if len(function.Blocks) == 0 {
```
samarth-aws, line 923: What if the function is a one-liner that calls another function? Is a function guaranteed to have a block if it's not empty?

analysis/dataflow/inter_procedural.go
```diff
@@ -650,6 +709,1040 @@ func UnwindCallStackToFunc(stack *CallStack, f *ssa.Function) *CallStack {
...
+		g.AnalyzerState.Logger.Debugf("Function %v has no blocks, skipping intra-procedural analysis", function)
+		// For functions with no blocks, create a minimal summary
+		summary.Constructed = true
+		summary.IsSound = true
+		return summary
+	}
+
+	if len(function.Blocks[0].Instrs) == 0 {
+		g.AnalyzerState.Logger.Debugf("Function %v has no instructions in first block, skipping intra-procedural analysis", function)
```
samarth-aws, line 932: Why do you only check the first block?

analysis/dataflow/inter_procedural.go
```diff
@@ -650,6 +709,1040 @@ func UnwindCallStackToFunc(stack *CallStack, f *ssa.Function) *CallStack {
...
+	function *ssa.Function,
+	summaryUnderCheck *SummaryGraph) (bool, string, map[*ssa.Function]*SummaryGraph) {
+
+	// Clone the summary-under-check to avoid modifying the original
+	Su := summaryUnderCheck
+	// Create a most-general summary (Sg) where every callee function has maximum dataflows
+	Sg := g.createMostGeneralSummary(function)
+
+	// Create a most-preserved summary (Sp) with no dataflows between callees
+	Sp := g.createMostPreservedSummary(function)
```
samarth-aws, line 834: I'm not sure why these are methods on the InterProceduralFlowGraph. Why not make them functions that return a SummaryGraph without using any inter-procedural information?

analysis/dataflow/inter_procedural.go
```diff
@@ -650,6 +709,1040 @@ func UnwindCallStackToFunc(stack *CallStack, f *ssa.Function) *CallStack {
...
+	// For each parameter, find all reachable target nodes using DFS
+	for _, paramNode := range summary.Params {
+		visited := make(map[GraphNode]bool)
+		reachableTargets := g.findReachableTargets(paramNode, targetNodes, visited)
+
+		// Record flows to reachable target nodes (excluding self-loops)
+		for _, target := range reachableTargets {
+			if target != paramNode { // Skip self-loops
+				flowKey := fmt.Sprintf("%s -> %s", paramNode.String(), target.String())
+				flows[flowKey] = true
```
samarth-aws, line 1088: Each node has an `Id` method that you can use to record flows. So `flows` could be `map[int32]int32`

analysis/dataflow/inter_procedural.go
```diff
@@ -650,6 +709,1040 @@ func UnwindCallStackToFunc(stack *CallStack, f *ssa.Function) *CallStack {
...
+
+	// Explore all outgoing edges
+	for destNode := range startNode.Out() {
+		// Skip already visited nodes to prevent cycles
+		if visited[destNode] {
+			continue
+		}
+
+		// Recursively explore this destination node
+		subTargets := g.findReachableTargets(destNode, targetNodes, visited)
```
samarth-aws, line 1118: nit: can you use a stack for DFS instead of recursion? we don't want to overflow the call stack and I don't think Go does many recursion optimizations

analysis/dataflow/inter_procedural.go
```diff
@@ -650,6 +709,1040 @@ func UnwindCallStackToFunc(stack *CallStack, f *ssa.Function) *CallStack {
...
+		return false
+	}
+	if ei1.Cond != nil && ei2.Cond != nil {
+		if ei1.Cond.Satisfiable != ei2.Cond.Satisfiable {
+			return false
+		}
+		if len(ei1.Cond.Conditions) != len(ei2.Cond.Conditions) {
+			return false
+		}
+		// For simplicity, we're not comparing the actual condition contents
```
samarth-aws, line 1192: Is it sound to not compare the condition contents?

samarth-aws:
What about soundness tests for your summary checking algorithm?

analysis/dataflow/inter_procedural.go
zcysky: I plan to have this feature in the next version with LLM summary generation. Currently it is just a placeholder. Will discuss it later.

analysis/dataflow/inter_procedural.go
zcysky: They are just used to mark the entry point for logging, I think they can be just removed.

analysis/dataflow/inter_procedural.go
zcysky: As current implements require assumption to callees, then it needs some kinds of inter-procedural information. Will discuss it later.

analysis/dataflow/inter_procedural.go
zcysky: See https://github.com/awslabs/ar-go-tools/pull/164/files/8dc5765de010686fd93f0284438b79fe2a4b29d6#r2248594481

analysis/dataflow/inter_procedural.go
zcysky: it's safe after discussion

analysis/dataflow/inter_procedural.go
zcysky: TODO: move that to the intra-procedural.go

PR #167, opened 2025-08-15
Fix nil panics

victornicolet:
+ This PR fixes many potential nil-panics. Those were identified by using `nilaway`, which now checks most of the analysis packages. Some packages have false positives and are not checked in CI for now.

One of the fixes (builtin type queried even in `defer (builtin)`) can actually be tested.

+ Fixes locations where an upgrade to the linter tools would throw warnings (formatting with non-constant format string)
+ Replaces use of `runtime.GOROOT` with the recommended approach to get the GOROOT. `runtime.GOROOT` will be deprecated in go 1.24.


internal/analysisutil/paths.go
```diff
@@ -15,15 +15,33 @@
 package analysisutil
 
 import (
+	"fmt"
+	"github.com/awslabs/ar-go-tools/internal/formatutil"
```
samarth-aws: nit: looks like goimports didn't catch this?

internal/analysisutil/paths.go
```diff
@@ -15,15 +15,33 @@
...
+var goRootAtInit = ""
+
+func init() {
+	cmd := exec.Command("go", "env", "GOROOT")
+	out, err := cmd.Output()
+	if err != nil {
+		return
+	}
+	goRootAtInit = strings.TrimSpace(string(out))
+	fmt.Fprint(os.Stderr, formatutil.Faint("Found GOROOT = %s", goRootAtInit)+"\n")
```
samarth-aws: nit: is this print necessary? I think we should only log the goroot in debug mode

samarth-aws (dismissed):

internal/analysisutil/paths.go
victornicolet: I suspect we might want this information at least initially when we test on different platforms. We don't have a debug flag available yet, so I don't know how to log only in debug mode in init?

samarth-aws (approved):

PR #170, opened 2025-11-20
MCP server with 1 tool: analyze dependencies.

victornicolet:
This PR adds a mcp server to argot. The only tool available in this first implementation is the `dependencies` tool of argot.


cmd/argot-mcp-server/main.go
```diff
@@ -0,0 +1,314 @@
...
+			},
+		},
+	}
+	sendResponse(req.ID, map[string]interface{}{"tools": tools})
+}
+
+func handleToolCall(req jsonRPCRequest) {
+	params, ok := req.Params.(map[string]interface{})
+	if !ok {
+		sendError(req.ID, -32602, "Invalid params")
```
samarth-aws, line 158: What are the error numbers for? Should they be constants?

cmd/argot-mcp-server/mcp_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"bufio"
+	"encoding/json"
+	"io"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestMCPServerDependenciesTool(t *testing.T) {
+	cmd := exec.Command("go", "run", "main.go")
```
samarth-aws, line 29: Can't the file be built the same way that we build files in our analysis tests? Why run a shell command?

cmd/argot-mcp-server/main.go
victornicolet: They are standard error codes for Json RPC: 
https://www.jsonrpc.org/specification#error_object 
I can document that and make them constants yes.

cmd/argot-mcp-server/mcp_test.go
victornicolet: What I want to do here is test the command, i.e. communication with a separate process the same way the MCP server will be being used.

victornicolet: Maybe it wasn't clear also, but it's running the mcp server here, not the file to be analyzed.

samarth-aws (approved):

PR #171, opened 2025-12-01
Cli as mcp

victornicolet:
This PR adds select CLI tools to the MCP server as well as a prompt. 
The goal is to allow the user to perform analysis using some agent, and also help with the task of generating dataflow summaries.
The `argot-mcp-server` now has 22 tools and one prompt.

The CLI tools are:
- argot_dataflow_summarize summarizes a set or a single function using the intra-procedural dataflow analysis,
- argot_dataflow_summary shows the summary of a function,
- argot_focused_mayalias show the mayalias relations of some SSA values when a function is in focus,
- argot_function_focus puts a function in focus, which enables usage of value inspections and other specific commands,
- argot_function_package prints information about the package of the function in focus,
- argot_function_unfocus  removes a function from the focus,
- argot_list_functions lists functions in the program,
- argot_load loads a program. The server always start without any program loaded. You should often start by loading a program,
- argot_members lists members of a package,
- argot_print_ast prints the AST of a function (the source code and the AST datastructure),
- argot_program_rebuild rebuilds the program from source (when a program is loaded),
- argot_run_pointer  runs the pointer analysis on the loaded program,
- argot_scan  scans for usages,
- argot_show_callees shows the calles of functions matching a regex argument,
- argot_show_callers shows the calles of functions matching a regex argument,
- argot_show_src shows the source code of a function,
- argot_show_ssa shows the SSA of a function,
- argot_show_ssa_instr  the SSA instruction matching a regex (requires a function in focus),
- argot_show_ssa_value  the SSA value matching a regex (requires a function in focus),
- argot_show_state show the state of the tool,
- argot_show_stats  show statistics about the loaded program,

Stateless tool:
- go_dependencies runs a dependency analysis.


cmd/argot/cli/analysis.go
```diff
@@ -45,55 +44,55 @@ import (
 // If state is nil, then it should print its definition on stdout
 
 // cmdShowSsa prints the SSA representation of all the function matching a given regex
-func cmdShowSsa(tt *term.Terminal, sess *session, command Command, withTest bool) bool {
+func cmdShowSsa(o Outputter, sess *Session, command Command, withTest bool) bool {
```
victornicolet, line 47: One of the main changes in the CLI is the use of `Outputter` instead of `*term.Terminal`. Outputter will be a wrapper around either the terminal (for the CLI) or a buffer (for the MCP server). When running for the MCP server, the terminal escape codes will be removed. 
This allows us to define CLI commands and MCP tools in one single function.

cmd/argot/cli/termpp.go
```diff
@@ -16,16 +16,239 @@ package cli
...
+// or standard writers. It automatically handles color formatting when writing to
+// a terminal and falls back to plain text when writing to regular io.Writers.
+//
+// The Outputter can operate in two modes:
+//   - Terminal mode: Uses a term.Terminal for colored output with escape sequences
+//   - Writer mode: Uses separate io.Writers for normal output and error output
+//
+// This abstraction allows the same code to work in both interactive terminal
+// environments and non-interactive contexts (like file output or pipes).
+type Outputter struct {
```
samarth-aws, line 34: I thought this would be an interface that the terminal and writer would implement? That way you don't need to branch on o.tt?

samarth-aws (dismissed):

cmd/argot/cli/termpp.go
victornicolet: Ah yes I guess it could be an interface.

cmd/argot-mcp-server/mcp_test.go
```diff
@@ -288,6 +288,75 @@ func checkListTool(t *testing.T, stdin io.WriteCloser, scanner *bufio.Scanner) {
...
+
+	// Test pagination: load argot program and list all functions
+	loadArgotReq := jsonRPCRequest{
+		JSONRPC: jsonRpcVersion,
+		ID:      8,
+		Method:  "tools/call",
+		Params: map[string]interface{}{
+			"name": "argot_load",
+			"arguments": map[string]interface{}{
+				"paths": []string{"./../../cmd/argot"},
```
samarth-aws, line 302: not urgent but this should probably be a separate file specific to the test so that you can actually check the length of the page results

samarth-aws (dismissed):

cmd/argot-mcp-server/mcp_test.go
victornicolet: For the pagination I need something that has enough functions in it to overflow a page -- so a test-size program will not do. We can have a separate test that counts functions too though.

samarth-aws (approved):

PR #173, opened 2025-12-10
Global as entrypoint

victornicolet:
Adds a feature: users can now specify that reading a global value can be a source of tainted data in the taint analysis.

### DIFF
--- a/analysis/taint/testdata/globals/main.go
+++ b/analysis/taint/testdata/globals/main.go
@@ -20,7 +20,8 @@ import (
 
 // TestTaintPropagatesThroughGlobal
 var x T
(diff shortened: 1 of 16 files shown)


analysis/taint/testdata/globals/main.go
```diff
@@ -20,6 +20,8 @@ import (
 
 // TestTaintPropagatesThroughGlobal
 var x T
+var globalSource1 T      // @Source(ex8)
```
samarth-aws, line 23: Can you add a test with a global pointer to make sure we track all aliases to globals?

samarth-aws (dismissed):

samarth-aws (approved):

PR #177, opened 2025-12-11
doc: add documentation that will appear on pkg.go.dev

victornicolet:
Argot has documentation in the doc folder, but that doesn't appear on pkg.go.dev. This adds documentation in the packages, so that there is basic information of what tools are available. This makes it easier for agents to discover the documentation.


cmd/static-commands/main.go
samarth-aws, line 1: Ideally this would be a separate commit

samarth-aws (approved):

cmd/static-commands/main.go
victornicolet: Yeah... in the process of documenting the commands I saw it and though it should go away.

